PR #599, opened 2022-02-25
Make populatefortest await on TransactionReceipt

shenkeyao:
Closes https://github.com/SpectrumXYZ/cape/issues/600.


wallet/src/bin/web_server.rs
```diff
@@ -928,24 +914,8 @@ mod tests {
...
-                .await
-                .unwrap()
-            {
-                minter_addr = Some(address);
-                break;
-            }
-        }
-        let minter = minter_addr.unwrap();
+        // Get the faucet address with non-zero balance of the native asset.
+        let minter: UserAddress = receipt.submitters[0].clone().into();
```
jbearer, line 918: 👍

wallet/src/web.rs
```diff
@@ -254,22 +270,41 @@ async fn populatefortest(req: tide::Request<WebState>) -> Result<tide::Response,
...
+            .add_user_key(faucet_key_pair.clone(), Default::default())
+            .await
+            .unwrap();
+    }
+    let faucet_addr = faucet_key_pair.address();
+    wallet.await_key_scan(&faucet_addr).await.unwrap();
 
     // Add a wrapped asset, and give it some nonzero balance.
-    let erc20_code = Erc20Code(EthereumAddr([1; 20]));
+    let mut erc20_code = Erc20Code(EthereumAddr([0; 20]));
```
jbearer, line 285: What if we sample this from entropy? Then we're almost guaranteed to not have a collision

jbearer (approved):

wallet/src/web.rs
shenkeyao: Good idea! I've modified this to use entropy.

PR #621, opened 2022-03-02
Integration of wallet with EQS

jbearer:


eqs/src/eth_polling.rs
```diff
@@ -100,12 +106,23 @@ impl EthPolling {
...
             .query_with_meta()
             .await
             .unwrap();
 
+        // don't refetch events against a block that was previously successfully processed
+        let mut block_cursor_advanced = false;
         for (filter, meta) in new_event {
+            let last_fetched_block = meta.block_number.as_u64();
+            let last_fetched_log_index = meta.log_index.as_u64();
+            // for some reason, we're seeing repeats in spite of the .from_block(last + 1) filter. Maybe ethers bug?
```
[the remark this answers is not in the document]
jbearer: I think that was intentional, when we switched from simple block-based indexing to (block, log_index) based. cc @nyospe to confirm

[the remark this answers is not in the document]
jbearer: I have updated the indexing logic once again with https://github.com/EspressoSystems/cape/pull/621/commits/e2b5fc4097ee8765bde97440057124d229c6ebc6

sveitser:
Just messing with some things to see if it makes helps the CI "re-attach".

eqs/src/eth_polling.rs
```diff
@@ -131,26 +158,28 @@ impl EthPolling {
...
-                        &state_lock.ledger_state.record_merkle_frontier,
-                    );
+                    let pending_commit =
+                        transitions.iter().cloned().chain(wraps).collect::<Vec<_>>();
+                    let output_record_commitments = pending_commit
+                        .iter()
+                        .flat_map(|txn| txn.output_commitments())
+                        .collect::<Vec<_>>();
+
+                    let mut merkle_tree = {
```
nyospe, line 176: 👍

eqs/src/eth_polling.rs
```diff
@@ -183,6 +212,16 @@ impl EthPolling {
                         })
                         .collect();
 
+                    let mut updated_state = self.query_result_state.write().await;
```
nyospe, line 215: I'm not a fan of this change. In the old code, the memos were being created before the QueryResultState saw any changes, which means eqs could continue serving queries. We could read the block_id in the same operation that we use to build the MerkleTree above, and avoid the early write lock. This has the added advantage of not changing the state at all if something goes wrong, but that's only semi-useful, since we don't really have any recoverable errors in this code segment. In fact, I'd actually prefer to see the Commit event created on the stack before the write() lock as well...

eqs/src/eth_polling.rs
```diff
@@ -295,45 +323,81 @@ impl EthPolling {
...
                     merkle_tree.push(rc.to_field_element());
                     let merkle_path = merkle_tree.get_leaf(uid).expect_ok().unwrap().1.path;
 
-                    // Process the memo
-                    let output = (memo.clone(), rc, uid, merkle_path.clone());
+                    // Generate commit event.
+                    let transition = CapeTransition::Faucet {
+                        ro: Box::new(ro.clone()),
+                    };
+                    let commit_event = LedgerEvent::Commit {
```
nyospe, line 344: 👍

eqs/src/query_result_state.rs
```diff
@@ -16,13 +16,21 @@ use seahorse::events::LedgerEvent;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, HashSet, VecDeque};
 
+/// The index of a single event in the Ethereum event stream.
+///
+/// Events are indexed using `(block_number, log_index)`, where `block_number` is the index of the
```
nyospe, line 21: 👍

sveitser (approved):
Don't understand everything here but what I understood LGTM.

wallet/src/backend.rs
```diff
@@ -225,12 +347,12 @@ impl<'a, Meta: Serialize + DeserializeOwned + Send> WalletBackend<'a, CapeLedger
             Some(ret) => Ok((ret, ())),
             None => {
                 let ret = self
-                    .mock_eqs
-                    .lock()
-                    .await
-                    .network()
-                    .nullifier_spent(nullifier);
-                nullifiers.insert(nullifier, ret);
+                    .get_eqs(format!("check_nullifier/{}", nullifier))
```
nyospe, line 350: Oof, so, yeah, this is the nullifier privacy leakage we talked about. I'm still mulling over ways to reduce the degree of leakage.

wallet/src/backend.rs
```diff
@@ -106,16 +116,38 @@ impl<'a, Meta: Serialize + DeserializeOwned + Send + Clone + PartialEq> CapeBack
...
         })
     }
 }
 
+impl<'a, Meta: Serialize + DeserializeOwned + Send> CapeBackend<'a, Meta> {
+    async fn get_eqs<T: DeserializeOwned>(
+        &self,
+        route: impl AsRef<str>,
+    ) -> Result<T, CapeWalletError> {
+        let mut res =
```
nyospe, line 135: Oh. Wow. That's an interesting pattern. It makes sense, though.

wallet/src/backend.rs
```diff
@@ -137,64 +169,154 @@ impl<'a, Meta: Serialize + DeserializeOwned + Send> WalletBackend<'a, CapeLedger
...
+            assets: Default::default(),
+            viewing_accounts: Default::default(),
+            freezing_accounts: Default::default(),
+            sending_accounts: Default::default(),
+        })
     }
 
     async fn subscribe(&self, from: EventIndex, to: Option<EventIndex>) -> Self::EventStream {
-        self.mock_eqs.lock().await.network().subscribe(from, to)
+        // Sleep at least 500ms between each request. This should try to match the EQS polling
```
nyospe, line 243: The EQS polling frequency is configurable... and may eventually be replaced with the ethers subscribe mechanism. I'm not saying capping this is a bad idea, but maybe something that can be configured, in case this is embedded or otherwise invoked in circumstances where half a second is too big a delay?

wallet/src/backend.rs
jbearer: I think this is actually OK (though admittedly it's a big footgun). The wallet only actually uses this endpoint in two places
1. Right before it spends a nullifier
2. When it's checking if a transaction has completed

1. is fine because it means you're about to publish a transaction which includes this nullifier anyways, and it has presumably been some time since you received the spent record, so it's hard for an attacker to correlate this nullifier with an earlier record commitment.

2. is tricky. It's fine for the wallet that published the transaction to check it this way. The other wallet that might do this is the receiver checking on the status of a transaction they expect to receive. This is quite useful in testing but probably shouldn't be exposed in any actual product, because there's not really any private way to do it and outside of a testing situation where you know ahead of time the nullifiers of the transaction you're expecting to receive, it's not that useful.

eqs/src/eth_polling.rs
jbearer: Yeah, fair point. I'll do it that way

wallet/src/backend.rs
jbearer: I love when markdown replaces "any number at all" with "1 + the last number I saw"

wallet/src/testing.rs
```diff
@@ -180,6 +205,43 @@ pub async fn get_burn_ammount<'a>(
...
 }
 
+pub fn rpc_url_for_test() -> Url {
+    match std::env::var("CAPE_WEB3_PROVIDER_URL") {
+        Ok(val) => val.parse().unwrap(),
+        Err(_) => "http://localhost:8545".parse().unwrap(),
+    }
+}
+
+pub async fn spawn_eqs(cape_address: Address) -> (Url, TempDir, JoinHandle<std::io::Result<()>>) {
```
nyospe, line 215: 👍

nyospe (commented):
Looks good, aside from the write() scope in eqs

nyospe (approved):
I like it. Sorry for the delay, this is good.

PR #658, opened 2022-03-03
Populating EQS data from CAPE contract

nyospe:


eqs/Cargo.toml
```diff
@@ -10,6 +10,7 @@ async-std = { version = "1.10.0", features = ["unstable", "attributes", "tokio1"
 atomic_store = { version="0.0.2", registry="translucence" }
 bincode = "1.3.3"
 cap-rust-sandbox = { path = "../contracts/rust" }
+commit = { git = "ssh://git@github.com/EspressoSystems/commit.git", rev = "f48cd52c59755eade0605111826eef3df6abdcf8" }
```
sveitser, line 13: Should use `https` url.

eqs/src/configuration.rs
```diff
@@ -27,9 +27,35 @@ pub struct EQSOptions {
...
 
+    /// URL for Ethers HTTP Provider
+    #[structopt(
+        long = "rpc_url",
+        env = "RPC_URL",
+        default_value = "http://localhost:8545"
+    )]
+    pub rpc_url: String,
+
+    /// Mnemonic for Ethers wallet
```
sveitser, line 38: @nyospe What will the EQS use the wallet for?

nyospe: It shouldn't, but the EthConnection setup needs it.

sveitser: Ok. I will try improve that.

eqs/Cargo.toml
nyospe: Oops, yeah, I think this getting added predated that change. I'll fix it.

eqs/src/eth_polling.rs
```diff
@@ -132,21 +123,16 @@ impl EthPolling {
                         .unwrap()
                         .0;
 
-                    //add transactions to QueryResultState pending commit
+                    let wraps = mem::take(&mut self.pending_commit_event);
```
nyospe, line 126: 👍

eqs/src/eth_polling.rs
```diff
@@ -250,6 +221,13 @@ impl EthPolling {
                     self.pending_commit_event.push(new_transition_wrap);
                 }
             }
+            let new_updated_block_height = meta.block_number.as_u64();
+            if new_updated_block_height > self.last_updated_block_height {
+                let updated_state = self.query_result_state.write().await;
+                // update the state block
+                // persist the state block updates (will be more fine grained in r3)
+                self.state_persistence.store_latest_state(&*updated_state);
```
nyospe, line 229: 👍

eqs/src/eth_polling.rs
```diff
@@ -250,6 +221,13 @@ impl EthPolling {
                     self.pending_commit_event.push(new_transition_wrap);
                 }
             }
+            let new_updated_block_height = meta.block_number.as_u64();
+            if new_updated_block_height > self.last_updated_block_height {
+                let updated_state = self.query_result_state.write().await;
+                // update the state block
```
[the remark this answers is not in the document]
kmchicoine: Ohhh so all of the writes happen here at once- that makes a lot more sense

eqs/src/eth_polling.rs
```diff
@@ -36,229 +39,215 @@ impl EthPolling {
...
-                    CAPEEvents::BlockCommittedFilter(_) => {
-                        let fetched_block_with_memos =
-                            fetch_cape_block(&self.connection, meta.transaction_hash)
-                                .await
-                                .unwrap()
-                                .unwrap();
+        for event in new_event {
+            let (filter, meta) = event.clone();
+            let new_updated_block_height = meta.block_number.as_u64();
+            if new_updated_block_height > self.last_updated_block_height {
```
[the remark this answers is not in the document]
jbearer: Not necessarily, those events are generated by separate transactions to different contract methods; they could easily end up in to different ETH blocks

eqs/src/eth_polling.rs
```diff
@@ -36,229 +39,215 @@ impl EthPolling {
...
+                            .ledger_state
+                            .record_merkle_commitment,
+                        &self
                             .query_result_state
                             .read()
                             .await
-                            .pending_commit_event
-                            .clone();
+                            .ledger_state
+                            .record_merkle_frontier,
```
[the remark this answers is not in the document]
jbearer: That should be documented clearly somewhere, that's one very big thing readers of this code won't have to think about (and if it ever becomes false, we'll want to know)

eqs/src/configuration.rs
```diff
@@ -44,7 +70,7 @@ pub struct EQSOptions {
 fn default_data_path() -> PathBuf {
     let mut data_dir = data_local_dir()
```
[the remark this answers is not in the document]
jbearer: I'll resolve this; since the wallet and EQS use storage for very different purposes (user-facing files vs internal-only persistent state) I think it's fine for the wallet to use `~` and the EQS to use `data_local_dir`, and then it doesn't matter if they have the same path or not

eqs/src/eth_polling.rs
```diff
@@ -19,6 +97,170 @@ impl EthPolling {
...
+                                .read()
+                                .await
+                                .ledger_state
+                                .record_merkle_frontier,
+                        );
+
+                        //add commitments to merkle tree
+                        let mut uids = Vec::new();
+                        let mut merkle_paths = Vec::new();
+                        if let Some(mut merkle_tree) = merkle_tree {
```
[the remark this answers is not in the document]
jbearer: Leaving this open, we don't have to fix this now, but I think we should at least log an error or something if this ever happens (I would even support panicking if this happens, at least for development, though we would want to be careful about not releasing that)

eqs/src/eth_polling.rs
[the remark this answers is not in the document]
nyospe: It isn't possible (unless I'm badly misinterpreting this) for the Erc20Deposited event to have a higher ETH block number than the BlockCommitted it is incorporated into. But is that what we want to use? The ETH block number? I thought we were using the commit event increasing height...

jbearer (approved):

PR #684, opened 2022-03-07
Add Wallet Endpoint for Getting the Last Used Keystore Path

bfish713:
The path is the location on disk of the most recently used wallet.  

This simply stores the serialized path to a file in .espresso/cape when we create or open a wallet.  Also it creates the $HOME/.espress/cape dir if it's not there.

closes #684


wallet/src/routes.rs
```diff
@@ -770,6 +798,10 @@ pub async fn get_records(wallet: &mut Option<Wallet>) -> Result<Vec<RecordInfo>,
     Ok(wallet.records().await.collect::<Vec<_>>())
 }
 
+pub async fn get_last_keystore() -> Result<PathBuf, tide::Error> {
```
bfish713, line 801: I could easily return another type here besides `PathBuf`.  is string or something else preferable

wallet/src/bin/web_server.rs
```diff
@@ -319,6 +319,89 @@ mod tests {
...
+    async fn test_lastusedkeystore() {
+        let server = TestServer::new().await;
+        let mnemonic = server.get::<String>("getmnemonic").await.unwrap();
+        println!("mnemonic: {}", mnemonic);
+        let password = "my-password";
+
+        // Should fail if no wallet exists.
+        server
+            .requires_wallet::<()>(&format!("openwallet/{}/path/{}", password, server.path()))
+            .await;
```
jbearer, line 333: Remove this

wallet/src/routes.rs
```diff
@@ -301,6 +304,36 @@ pub fn wallet_error(source: CapeWalletError) -> tide::Error {
...
+        .write_all(&bincode::serialize(&wallet_path).unwrap())
+        .await?)
+}
+pub async fn read_last_path() -> Result<PathBuf, tide::Error> {
+    let mut path = get_home_path().unwrap();
+    path.push(".espresso/cape/last_wallet_path");
+    let mut file = File::open(&path).await?;
+    let mut bytes = Vec::new();
+    file.read_to_end(&mut bytes).await?;
+    Ok(bincode::deserialize(&bytes).unwrap())
```
jbearer, line 335: I'd prefer not to unwrap here; since we're reading data from the user's file system it's very possible that it doesn't deserialize correctly. We should just return an error in this case (note that unwrapping when we serialize, in the previous function, should be fine since we've validated/constructed the data ourselves)

wallet/src/routes.rs
jbearer: Also on the subject of return type, I think we should return `Option<something>` here, returning `None` (which turns into JSON `null`) when there is no last opened path

jbearer (changes requested):

wallet/src/routes.rs
bfish713: Here is the exact result from the curl, so it looks like it will be human readable
```
curl http://0.0.0.0:60000/lastusedkeystore
"/tmp/test_cape_wallet.pOvokKcH0DNA"%  
```

bfish713:
I still need to figure out why this is failing on CI

wallet/src/bin/web_server.rs
```diff
@@ -327,17 +328,27 @@ mod tests {
...
         let password = "my-password";
 
-        // Should fail if no wallet exists.
-        server
-            .requires_wallet::<()>(&format!("openwallet/{}/path/{}", password, server.path()))
-            .await;
+        // clean up from any previous test run on the current machine
+        let mut storage_path = get_home_path().unwrap();
+        storage_path.push(".espresso/cape/last_wallet_path");
+        remove_file(storage_path).unwrap();
```
jbearer, line 334: Actually, can we have an env var to override the default `.espresso` path, and then use a tempdir for this test? That might also be pretty useful for Plaza as they're developing/testing

bfish713: Updated the code based on this, using /tmp for the test

wallet/src/routes.rs
```diff
@@ -313,19 +313,33 @@ pub fn get_home_path() -> Result<PathBuf, tide::Error> {
     Ok(PathBuf::from(home))
 }
 
-pub async fn write_path(wallet_path: &Path) -> Result<(), tide::Error> {
+pub async fn defult_storage_path() -> Result<PathBuf, tide::Error> {
```
jbearer, line 316: `default`

wallet/src/bin/web_server.rs
```diff
@@ -328,9 +330,10 @@ mod tests {
         println!("mnemonic: {}", mnemonic);
         let password = "my-password";
 
+        env::set_var("PATH_STORAGE", "/tmp");
```
jbearer, line 333: Slightly better to use
```suggestion
        let dir = TempDir::new("wallet_last_used_test").unwrap();
        env::set_var("PATH_STORAGE", &format!("{:?}", dir));
```
but I won't let that block the review

bfish713:
The test fails because the other tests are also creating and opening wallets in parallel.  Even though the `lastusedkeystore` endpoint is behaving correctly the last used keystore is likely to be coming from another test.  Setting the env var only helps slightly, but since it applies to all tests, the other tests are still writing to the last used wallet file.

My ideas for a fix are: 
- Using command line param to specify the path, this way it only applies to the one test/one instance of the web server.
- Using RW lock in the tests to prevent any other test from creating/opening a wallet while the lastusedkeystore test runs

wallet/src/bin/web_server.rs
```diff
@@ -146,7 +152,7 @@ mod tests {
     }
 
     impl TestServer {
-        async fn new() -> Self {
+        async fn new(path_storage: Option<PathBuf>) -> Self {
```
jbearer, line 155: Bit cleaner if we just make this a TempDir, generated in `new()`, instead of passing it in from the test. Not a big deal though

jbearer (approved):

wallet/src/bin/web_server.rs
bfish713: Oh yeah good point.  For some reason I was thinking that would mean all test write to the same place, but that's not how TempDir works.  I'll make this change before merging since it's simple enough.

Issue #455, opened 2022-02-04
EQS API: obtain an AAP transaction by hash


jbearer:
* `get_transaction_by_hash(hash: Commitment<CapeTransition>) -> CommittedCapeTransition`
    - Required for wallet transaction tracking, which is especially important for synchronizing between different wallets to avoid race conditions in tests
    - Should fail with a reasonable error message if `hash` does not identify a previously committed transaction
    - `CommittedCapeTransition` is as defined in #247

philippecamacho:
Is not it a duplicate of #247? @nyospe @kmchicoine

nyospe:
> 

It is not, in fact, a duplicate. The two endpoints have a different index. This one lets the wallet find a transaction it submitted, if it has been committed, without yet having processed the event for its commit.

nyospe:
This is the one I need to add a new TaggedBase64 wrapper type for.

PR #693, opened 2022-03-08
get_transaction_by_hash added to routes

nyospe:
Creates the endpoint for https://github.com/EspressoSystems/cape/issues/455

### DIFF
--- a/eqs/api/api.toml
+++ b/eqs/api/api.toml
@@ -56,4 +56,4 @@ DOC = "Returns the committed transaction, if any, at the specified location, wit
 PATH = [ "get_transaction/:hash" ]
 ":hash" = "TaggedBase64"
 DOC = "Returns the committed transaction, if any, with the specified hash, with output indexes."
-ERROR_hash = "A valid commitment hash is required. Commitment begin with NUL~."
+ERROR_hash = "A valid commitment hash is required. Commitment begin with CMTMNT_CAPE_TRNSTN~."
--- a/eqs/api/api.toml
+++ b/eqs/api/api.toml
@@ -52,8 +52,8 @@ PATH = [ "get_transaction/:block_id/:txn_id" ]
 ":txn_id" = "Integer"
 DOC = "Returns the committed transaction, if any, at the specified location, with output indexes."
 
-# [route.get_transaction_by_hash]
-# PATH = [ "get_transaction/:hash" ]
-# ":hash" = "TaggedBase64"
-# DOC = "Returns the committed transaction, if any, with the specified hash, with output indexes."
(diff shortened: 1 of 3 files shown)


eqs/src/routes.rs
```diff
@@ -148,19 +148,21 @@ pub async fn get_transaction(
...
+    bindings: &HashMap<String, RouteBinding>,
+    query_result_state: &QueryResultState,
+) -> Result<Option<CommittedCapeTransition>, tide::Error> {
+    if let Some(txn_id) = query_result_state.transaction_id_by_hash.get(
+        &bindings[":hash"]
+            .value
+            .to::<CommitmentToCapeTransition>()?
+            .0,
+    ) {
+        Ok(query_result_state.transaction_by_id.get(txn_id).cloned())
```
jbearer, line 161: Maybe it should be an error if the hash exists but its id is invalid? This violates internal consistency assumptions right?

eqs/api/api.toml
```diff
@@ -52,8 +52,8 @@ PATH = [ "get_transaction/:block_id/:txn_id" ]
...
-# [route.get_transaction_by_hash]
-# PATH = [ "get_transaction/:hash" ]
-# ":hash" = "TaggedBase64"
-# DOC = "Returns the committed transaction, if any, with the specified hash, with output indexes."
-# ERROR_hash = "A valid commitment hash is required. Commitment begin with NUL~."
+[route.get_transaction_by_hash]
+PATH = [ "get_transaction/:hash" ]
+":hash" = "TaggedBase64"
+DOC = "Returns the committed transaction, if any, with the specified hash, with output indexes."
+ERROR_hash = "A valid commitment hash is required. Commitment begin with NUL~."
```
jbearer, line 59: It doesn't begin with `NUL~`.

nyospe: Oops.

eqs/src/routes.rs
nyospe: Making that an error case, but I'm wondering if I shouldn't also switch to forming the tide::Response directly for this endpoint, so that we can have a (non-error) `tide::Response::new(tide::StatusCode::NotFound)` for the not found case.

nyospe: Would it make sense for the net::server::response function to automate that for `Result<Option<BodyType>>`?

jbearer: I think different things make sense for different endpoints. In my pretty uninformed opinion, `tide::StatusCode::NotFound` feels more error-ish, like you asked for something you shouldn't have. Whereas for something that is explicitly optional, returning a `JSON` object that might be `null` seems easier to work with in JS, and that corresponds to returning `None` with status 200.

jbearer (approved):

PR #694, opened 2022-03-08
Added signal handler for SIGINT and SIGTERM

pictographer:
Splitting #547 into smaller pull requests. This is the basic signal handler. Another pull request will add a communication channel that tells the web server to stop accepting new requests and exit gracefully.


address_book/Cargo.toml
```diff
@@ -16,5 +17,6 @@ net = { git = "https://github.com/EspressoSystems/net.git" }
 once_cell = "1"
 rand_chacha = "0.3.1"
 serde = { version = "1.0", features = ["derive"] }
+signal-hook = "0.3.13"
```
VictorKoenders, line 20: `signal-hook` explicitly mentions:

> If you are looking for integration with an asynchronous runtime take a look at one of the following adapter crates:
> [signal-hook-async-std](https://docs.rs/signal-hook-async-std) for async-std support

address_book/src/signal.rs
```diff
@@ -0,0 +1,23 @@
...
+use async_std::task::spawn;
+use async_std::task::JoinHandle;
+use signal_hook::{consts::SIGINT, consts::SIGTERM, iterator::Signals};
+use std::process;
+
+/// Spawn a thread that waits for SIGTERM. If SIGTERM is received,
+/// the application exits with exit status 1.
+pub async fn init_sig_handler() -> Result<JoinHandle<()>, ()> {
+    let mut signals = Signals::new(&[SIGINT, SIGTERM]).unwrap();
+    Ok(spawn(async move {
```
VictorKoenders, line 17: Because this is a blocking thread you should use [spawn_blocking](https://docs.rs/async-std/latest/async_std/task/fn.spawn_blocking.html)

address_book/Cargo.toml
pictographer: Ah, thanks! Is this better?

jbearer (approved):
LGTM

VictorKoenders (approved):
LGTM

PR #701, opened 2022-03-08
Use plain base64 for UI-constructed parameters.

jbearer:
There is currently no JS/WASM library for constructing tagged base
64. This changes URL parameters which the UI is expected to construct
from scratch, that were using tagged base 64, to use plain base 64
instead, which is easier to construct in JS.

We should change these back to tagged base 64 if and when we have a
compatible JS library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1123,6 +1123,7 @@ dependencies = [
  "address_book",
  "async-std",
  "async-trait",
+ "base64 0.13.0",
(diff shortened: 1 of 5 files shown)


tri-joe:
LGTM, although we should fix this later, I set up https://github.com/EspressoSystems/cape/issues/712 to track it

tri-joe (approved):

PR #704, opened 2022-03-09
persistence support for address_book

nyospe:


sveitser:
@nyospe For the demo we would like to run at least 2 instances of the address book (and relayer, EQS) in 2 availability zones for high availability. At the moment if we were to run 2 instances would each instance get some of the writes and we'd end up with 2 different address books?

Would it be desirable to make the address book service stateless by persisting each key/value pair individually in EFS (or even S3, then there's a global single source of truth) and look up the keys on each request?

cc @Ancient123 @pictographer

nyospe:
Looking at the address book, the answer to the first question is clearly, yes. Even if we want to share that state, the persistence layer is not the mechanism to do this. Backing it against a distributed (NoSQL) database would be more suitable for that, but that wouldn't work if the reads were still going to in-memory storage; it would require reads to go against the database as well. The alternative is to have address books synchronizing and distributing state.

Honestly, given that, this would be more appropriately handled by a thin Node service on top of something like Dynamo.

sveitser:
I created
- https://github.com/EspressoSystems/cape/issues/716
for further discussion on this issue.

nyospe:
Obsolete after https://github.com/EspressoSystems/cape/pull/718

PR #705, opened 2022-03-09
impl batch inversion

zhenfeizhang:
gas cost for pcs_info

- before optimization 451956
- after batch inversion 451477
- after moving everything inside assembly: 450007

### DIFF
--- a/contracts/contracts/libraries/PolynomialEval.sol
+++ b/contracts/contracts/libraries/PolynomialEval.sol
@@ -166,7 +166,9 @@ library PolynomialEval {
         uint256 divisor;
         uint256 tmp;
         uint256 vanishEvalDivN = self.sizeInv;
+        uint256 divisorProd;
         uint256[] memory localDomainElements = domainElements(self, length);
+        uint256[] memory buffer = new uint256[](length);
 
         // vanish_eval_div_n = (zeta^n-1)/n
         assembly {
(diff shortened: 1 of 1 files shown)


contracts/contracts/libraries/PolynomialEval.sol
```diff
@@ -180,18 +182,38 @@ library PolynomialEval {
...
+                ithLagrange := mulmod(vanishEvalDivN, tmp, p)
+            }
             // compute 1/(zeta - g^i)
-            divisor = BN254.invert(divisor);
+            divisor = divisorProd;
+            for (uint256 j = 0; j < length; j++) {
+                if (i != j) {
+                    divisor = mulmod(divisor, buffer[j], p);
+                }
+            }
```
alxiong, line 216: we can turn this into full assembly easily so that we can have a single assembly block in this for-loop, wdyt (you can decide after comparing the gas cost, I agree that right now your code is more readable)?

```solidity
ithDivsiorInv := divisiorProd
for { let j := 0 } lt(j, length) { i := add(i, 1) }
{
    if not(eq(i, j)) { 
        ithDivisor := mload(add(buffer, add(0x20, mul(j, 0x20))))
        ithDivisorInv := mulmod(ithDivisorInv, ithDivisor, p) 
    }
}
```

zhenfeizhang: good idea. we can put more comment in the code and use your version

zhenfeizhang:
I moved almost the whole function inside assembly, and that gives us a bigger saving than batch inversion. LOL.
We should consider moving other functions to assembly for deliverable 3 cc @alxiong @philippecamacho

contracts/contracts/libraries/PolynomialEval.sol
```diff
@@ -163,63 +163,91 @@ library PolynomialEval {
...
+                // now we compute vanish_eval_div_n * g^i / (zeta - g^i) via
+                // vanish_eval_div_n * g^i * divisorProd * \prod_{j!=i} (zeta - g^j)
+                ithLagrange := mulmod(ithLagrange, divisorProd, p)
+                for {
+                    let j := 0
+                } lt(j, length) {
+                    j := add(j, 1)
+                } {
+                    // here we cannot use not(eq(i, j)): not(x) will flip all bits of x
+                    if eq(eq(i, j), 0) {
```
alxiong, line 241:
```suggestion
                    if iszero(eq(i, j)) {
```

contracts/contracts/libraries/PolynomialEval.sol
```diff
@@ -163,63 +163,91 @@ library PolynomialEval {
...
+            //
+            // and then each 1 / (zeta - g^i) can be computed via (length - 1)
+            // multiplications:
+            //
+            //      1 / (zeta - g^i) = divisorProd * \prod_{j!=i} (zeta - g^j)
+            //
+            // In total this takes n(n-1) multiplications and 1 inversion,
+            // instead of doing n inversions.
+            //
+            // credit: https://github.com/EspressoSystems/cape/issues/689
```
alxiong, line 197:
```suggestion
```
People can git blame to see the related issue. no need for further comment.

alxiong (commented):
Generally LGTM, some minor comment

> We should consider moving other functions to assembly for deliverable 3 

We probably could, but still need to strike a balance in readability/auditability and gas efficiency. I'd say most of Plonk verifier code can be in assembly.

alxiong (approved):

PR #706, opened 2022-03-09
Remove flaky contract deployment tests

### LINKED ISSUE #501: Flaky test `test_basic_contract_transaction`
https://github.com/SpectrumXYZ/cape/runs/5183173124?check_suite_focus=true

```
[2022-02-14 11:58:06] failures:
[2022-02-14 11:58:06] 
[2022-02-14 11:58:06] ---- test_basic_contract_transaction stdout ----
error: test failed, to rerun pass '-p cap-rust-sandbox --test ethereum_test'
[2022-02-14 11:58:06] Sent funding tx to deployer
[2022-02-14 11:58:06] thread 'test_basic_contract_transaction' panicked at 'called `Result::unwrap()` on an `Err` value: AbiError(DecodingError(InvalidName("please ensure the contract and method you're calling exist! failed to decode empty bytes. if you're using jsonrpc this is likely due to jsonrpc returning `0x` in case contract or method don't exist")))', contracts/rust/tests/ethereum_test.rs:20:53
```

sveitser:
These are tests from the very beginning when we occasionally had
problems with simple contract deployments. They are now more harmful
than useful because they occasionally make the CI runs fail and cause
confusion.

Close #501


alxiong (changes requested):
should also remove https://github.com/EspressoSystems/cape/blob/maint/t501-remove-flaky-tests/contracts/test/greeter.spec.ts

Do you also want remove `SimpleToken.sol` and related test or deploy codes?

sveitser:
I thought it wasn't a bad idea to have an example of a very simple test and contract in the repo in there that's why I left it there.

sveitser:
Regarding `SimpleToken`: looks to me it is used in tests that need an ERC20 token.

alxiong (approved):

PR #711, opened 2022-03-09
Load faucet memo and update merkle tree when EQS parses the faucet event

### LINKED ISSUE #683: Load faucet memo and update merkle tree at EQS initialization

philippecamacho:
Closes #683.

### DIFF
--- a/eqs/src/eth_polling.rs
+++ b/eqs/src/eth_polling.rs
@@ -20,7 +20,10 @@ use cap_rust_sandbox::{
 use commit::Committable;
 use core::mem;
 use ethers::abi::AbiDecode;
+use jf_cap::structs::{ReceiverMemo, RecordCommitment};
 use jf_cap::{structs::RecordOpening, MerkleTree, TransactionNote};
+use rand_chacha::rand_core::SeedableRng;
+use rand_chacha::ChaChaRng;
 use reef::traits::{Block, Transaction};
 use seahorse::events::LedgerEvent;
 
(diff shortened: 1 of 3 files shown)


eqs/src/eth_polling.rs
```diff
@@ -277,8 +281,58 @@ impl EthPolling {
...
+                    let mut rng = ChaChaRng::from_entropy();
+                    let memo = ReceiverMemo::from_ro(&mut rng, &ro, &[]).unwrap();
+
+                    // Update the state
+                    let faucet_transition = Faucet {
+                        memo: Box::new(memo.clone()),
+                        ro: Box::new(ro),
+                    };
+
+                    self.pending_commit_event.push(faucet_transition);
```
nyospe, line 302: I think we we have a small inconsistency here, in the sense that we use pending_commit_event as an accumulator between events that get made available to the wallet (e.g. erc20_deposited), but this is actually something that will be used to generate a `LedgerEvent::Memos` **now**. I think, from the wallet's perspective, this event is not visible.

Remove this line and the block above, we'll add them back if we add a raw_events endpoint. 

Otherwise, this looks good.

nyospe (changes requested):
I think, if you get rid of the pending_commit_event.push, this is all good.

eqs/src/eth_polling.rs
philippecamacho: Right, I was unsure about adding this *transition*. The code has been removed in 6dce87f.

nyospe (approved):
LGTM

jbearer (approved):

PR #718, opened 2022-03-09
Basic stateless address book with file store

### LINKED ISSUE #504: Address book: add persistence

### LINKED ISSUE #716: Make address book stateless
Instead of having an in memory hash map the address book should save the entries to a persistent store (file, S3, database, ...) and lookup the entries from there.

sveitser:
### TODO

- [x] Configuration
- [x] Use temporary directory for tests
- [x] Prevent concurrent writes (if necessary)
- [x] Clean up after tests (not working for wallet tests at the moment, see 2bee5bdfb07f1590b3e6a16d56c2be9212590371, edit: fixed in 2c39aa11bb9247319ab5ef35ab089e0841bbc748)

Close #716 
Close #504


address_book/src/lib.rs
```diff
@@ -32,15 +29,48 @@ static LOGGING: Lazy<()> = Lazy::new(|| unsafe {
...
+        Self { dir }
+    }
+    fn path(&self, address: UserAddress) -> PathBuf {
+        let as_hex = hex::encode(bincode::serialize(&address).unwrap());
+        self.dir.join(format!("{}.bin", as_hex))
+    }
+}
+
+impl Store for FileStore {
+    fn save(&self, address: UserAddress, pubkey: UserPubKey) -> Result<(), std::io::Error> {
```
nyospe, line 53: From the documentation, I'm not seeing anything guaranteeing this is incapable of a mid-write flush with subsequent read collision under NFS-over-EFS.

It would, however, be guaranteed if you did the fs::write to a path with a temp suffix (preferably with a random component) and then did a fs::rename after the write.

sveitser: Thanks. Yeah so far I haven't been able to figure out what kind of behavior we would see in practice. I think we're unlikely to run into any issues for the demo but I don't mind adding the tempfile/rename solution. Although one wonders if we shouldn't just use S3.

sveitser: This may be relevant https://stackoverflow.com/a/11402827 but no link to any documentation. It suggests that if we stay below the block size the files should remain consistent.

nyospe: Sure, or Dynamo, but both of those require using an SDK, with the learning and deployment overhead that comes with that, both are AWS lock-in, and testing isn't quite as transparently symmetrical. This solution has the benefit of being both simple and trivial to test locally, and it works on any cloud service that provides a distributed filesystem that can map to a unix-like file system (e.g. Azure or Google), which may not matter that much, but is still nice.

I know from prior experience that EFS updates are fast enough that it's hard to calibrate the delay across nodes vs delay from other factors (network access to the node access point, etc.), and that the propagation time for remotely created files was small enough that it got buried in the network noise for a native (C++) testing framework pinging its counterpart over a direct network socket.

nyospe: Oh, right, I forgot about the NFS negotiated block message. You actually get to set the block size when you implement your NFS adaptor, but there's a bunch of stuff about unstable writes, and I think AWS is using the v4 standard. It's been a long time since I've dealt with this, and nothing had actually implemented v4 last time I had to muck with the guts of NFS. I'm making a note to myself to do some reading when I've got some time.

For what it's worth, this is the AWS documentation: https://docs.aws.amazon.com/efs/latest/ug/mounting-fs-old.html#mounting-fs-nfs-info and current NFS documentation: https://datatracker.ietf.org/doc/html/rfc3530.txt (4.0) and https://datatracker.ietf.org/doc/html/rfc5661 (4.1)

The default mount is in NFS 4.1 mode. It looks like their default configured block size is actually 1MB (https://docs.aws.amazon.com/efs/latest/ug/efs-mount-helper.html#efs-mount-helper-how - see the rsize/wsize settings). So, yeah, even without the rename, this _should_ be fine. The rename pattern is, however, best practice for a reason, and I'd prefer its use.

address_book/src/lib.rs
```diff
@@ -32,31 +31,101 @@ static LOGGING: Lazy<()> = Lazy::new(|| unsafe {
...
-    map: Arc<RwLock<HashMap<UserAddress, UserPubKey>>>,
+    store: FileStore,
 }
 
 pub fn address_book_port() -> String {
     std::env::var("PORT").unwrap_or_else(|_| DEFAULT_PORT.to_string())
 }
 
+pub fn cape_data_path() -> PathBuf {
+    dirs::data_local_dir()
```
sveitser, line 98: @nyospe Took this from PR #504 (with a bit of rewriting).

address_book/src/lib.rs
```diff
@@ -32,31 +31,101 @@ static LOGGING: Lazy<()> = Lazy::new(|| unsafe {
     tide::log::with_level(LOG_LEVEL);
 });
 
+// TODO: remove test store after tests, add TestFileStore type that implements Drop?
```
sveitser, line 34: @nyospe I tried to do this but it turned out pretty messy because I didn't find a working way to organize the traits and structs.

Do you know how I could have a version of the Store for tests that implements `Drop` and deletes the store path when inside `drop`? Or any other way to clean up more or less reliably after the tests. At least the files should not remain if the tests fail.

address_book/tests/tests.rs
```diff
@@ -21,7 +23,7 @@ const NOT_FOUND_COUNT: u64 = 100;
 #[async_std::test]
 async fn round_trip() {
     // TODO !corbett find an unused port rather than assuming 50078 is free.
-    init_web_server(LevelFilter::Error)
+    init_web_server(LevelFilter::Error, env::temp_dir())
```
sveitser, line 26: Just realized `env::temp_dir()` doesn't do what I think it does. It's the systems temp dir. Will fix that.

address_book/src/lib.rs
sveitser: This is already built into `tempdir::TempDir`. I'm just not using it the right way at the moment.

address_book/src/lib.rs
```diff
@@ -29,69 +35,131 @@ static LOGGING: Lazy<()> = Lazy::new(|| unsafe {
...
-    fn load(&self, address: UserAddress) -> Option<UserPubKey> {
+    async fn load(&self, address: &UserAddress) -> Option<UserPubKey> {
         match fs::read(self.path(address)) {
             Ok(bytes) => Some(bincode::deserialize(&bytes).unwrap()),
             Err(_) => None,
         }
     }
 }
 
+/// Non-persistent in-memory store. Suitable for testing only.
```
nyospe, line 93: But why, though? Wouldn't using the same filesystem store code, but with a random tempdir that gets created on init and deleted on scope exit be better?

address_book/src/lib.rs
nyospe: 👍

address_book/src/lib.rs
sveitser: Yes.

sveitser: https://github.com/EspressoSystems/cape/pull/718/commits/1038703b7866671b5b45f38fc4829645e93b85ef

nyospe (approved):
LGTM

Issue #596, opened 2022-02-25
Implement a `send` variant that aggregates from all addresses


jbearer:
Depends on SpectrumXYZ/seahorse#26

PR #723, opened 2022-03-10
Implement a send variant that aggregates from all addresses

shenkeyao:
Closes https://github.com/EspressoSystems/cape/issues/596.


jbearer (approved):
LGTM

PR #727, opened 2022-03-10
Support named keystores with an endpoint to list them.

jbearer:
The bulk if this change is reworking the way storage is configured,
so that there is now one parameter for the root storage directory
for all wallets. This replaces `~/.espresso/cape/wallet` if given.
This directory is then used to store both the last used wallet path
(`$CAPE_WALLET_STORAGE/last_wallet_path`) and named keystores
(`$CAPE_WALLET_STORAGE/keystores/`).

We then extend the `newwallet` and `openwallet` endpoints to take a
:name instead of a :path, which automatically stores the created
wallet in `$CAPE_WALLET_STORAGE/keystores/$NAME`. We add an endpoint
`listkeystores` to list all of the keystore names in the keystores
directory.

### DIFF
--- a/wallet/api/api.toml
+++ b/wallet/api/api.toml
@@ -38,26 +38,41 @@ PATH = ["getmnemonic"]
 DOC = "Generate a random mnemonic phrase."
 
 [route.newwallet]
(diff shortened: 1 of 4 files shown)


bfish713 (approved):
Makes sense to me

PR #730, opened 2022-03-11
Use nix based docker base image

sveitser:
### Motivation

1. No need for patchelf. Base image inputs are tracked with nix.
2. Make it possible to build the image locally. Should make it easier to put together the remaining images we need.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,11 +98,10 @@ jobs:
           name: wallet-webserver
           path: wallet-checkout/target/release/
 
-      - name: Set linker/loader to not be nix
-        run: patchelf --set-interpreter /lib64/ld-linux-x86-64.so.2 wallet-checkout/target/release/web_server
-
       - uses: docker/setup-buildx-action@v1
         name: Setup Docker BuildKit (buildx)
+        with:
+          driver: docker
 
       - uses: docker/login-action@v1
         name: Login to Github Container Repo
@@ -117,6 +116,9 @@ jobs:
         id: meta
         with:
           images: ghcr.io/espressosystems/cape/wallet
+
+      - name: Generate base image
+        run: cd wallet-checkout && bin/build-docker-base
 
       - uses: docker/build-push-action@v2
         name: Build and Push Docker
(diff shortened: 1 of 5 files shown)


sveitser:
Building the base image works on the CI bu the final "Build and Push Docker" step fails. I think it's because it's trying to fetch metadata about the base image from dockerhub and that image doesn't exist.

https://github.com/EspressoSystems/cape/runs/5511464159?check_suite_focus=true

sveitser:
Possibly relevant: https://github.com/docker/build-push-action/issues/425#issuecomment-892807549

.github/workflows/build.yml
```diff
@@ -118,6 +117,9 @@ jobs:
         with:
           images: ghcr.io/espressosystems/cape/wallet
 
+      - name: Generate base image
+        run: nix-shell --run build-docker-base
```
sveitser, line 121: This is a bit wasteful and we don't really need to activate the nix shell but I'm not the `nix build ...` command works without it.

sveitser: Will give it a go when I get around to it.

Ancient123 (approved):

DieracDelta (approved):

.github/workflows/build.yml
```diff
@@ -118,6 +117,9 @@ jobs:
         with:
           images: ghcr.io/espressosystems/cape/wallet
 
+      - name: Generate base image
+        run: bash bin/build-docker-base
```
Ancient123, line 121: This is gonna actually be `wallet-checkout/bin/build-docker-base` for weird permissions reasons.

Ancient123 (changes requested):

.github/workflows/build.yml
sveitser: @Ancient123 seems to work now.

.github/workflows/build.yml
sveitser: This is done. The docker workflow just takes seconds now.

Ancient123 (approved):

PR #731, opened 2022-03-11
Faucet setup: encryption/public key in calldata

### LINKED ISSUE #726: Emit public key as part of `FaucetInitialized` Event.
https://translucence.zulipchat.com/#narrow/stream/311688-CAPE

sveitser:
## Notes
- The hardhat deploy script is probably not correct yet.
- I'm mainly wondering if the rest is going in the right direction.

Close #726

### DIFF
--- a/contracts/rust/src/bin/faucet.rs
+++ b/contracts/rust/src/bin/faucet.rs
@@ -1,12 +1,28 @@
 use ark_std::str::FromStr;
+use itertools::Itertools;
 use jf_cap::keys::UserPubKey;
+
+// Fp256 "(2DCA81140764685EBFAC3C684E0FF0DB3500A853AB3EE0C966D463AC547BE39A)"
+// => 2DCA81140764685EBFAC3C684E0FF0DB3500A853AB3EE0C966D463AC547BE39A
+fn disp_hex<T: ToString>(x: T) -> String {
+    let vec = x.to_string().chars().collect_vec();
+    vec[8..vec.len() - 2].iter().collect()
+}
 
 fn main() {
     let result = "USERPUBKEY~muN7VKxj1GbJ4D6rU6gANdvwD05oPKy_XmhkBxSByq0gAAAAAAAAAIRN-Rik8czFiToI8Ft5fsIf9HAEtWHDsOHh-ZBJZl1KxQ";
-    let pk = UserPubKey::from_str(result).unwrap_or_default();
-    ark_std::eprintln!(
-        "x: {}, y: {}",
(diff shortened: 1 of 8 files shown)


philippecamacho:
As we only use a single address/public key for the faucet which is known in advance, why not hardcoding it into the rust code as well? This way we do not need to use call data and make type conversions?

WDYT @sveitser @alxiong @jbearer ?

sveitser:
It's useful for testing to be able to set it programatically.

contracts/rust/src/cape/faucet.rs
```diff
@@ -34,26 +34,34 @@ async fn test_faucet() -> Result<()> {
     assert!(!contract.faucet_initialized().call().await?);
     assert_eq!(contract.deployer().call().await?, deployer.address());
 
+    let pub_key_bytes = bincode::serialize(&faucet_manager.pub_key()).unwrap();
```
jbearer, line 37: I think this is fine because it's just a test and the contract itself doesn't care what the pub key bytes are, but in general we should always be consistent (either always use `TaggedBase64` or always use `bincode` bytes) so that consumers of the event (e.g. EQS) know what format to expect

eqs/src/eth_polling.rs
```diff
@@ -289,10 +290,12 @@ impl EthPolling {
                 }
                 CAPEEvents::FaucetInitializedFilter(filter_data) => {
                     let ro_bytes = filter_data.ro_bytes;
+                    let pub_key: UserPubKey =
+                        bincode::deserialize(&filter_data.faucet_manager_pub_key.to_vec()).unwrap();
```
jbearer, line 294: This won't work with the deployment script which uses UTF-8 bytes from the TaggedBase64 serilization.

jbearer (changes requested):

sveitser:
I'm off for today. If it's okay feel free to merge or otherwise make changes on the branch as desired.

contracts/rust/src/bin/faucet.rs
```diff
@@ -1,12 +1,28 @@
...
-        pk.address().internal().y
+    let pub_key = UserPubKey::from_str(result).unwrap_or_default();
+
+    println!(
+        r#"
+const pubKey = "{}";
+const faucetManagerAddress = {{
+  x: BigNumber.from("0x{}"),
+  y: BigNumber.from("0x{}"),
+}};
```
jbearer, line 22: Beautiful

jbearer (approved):
Looks good, thanks a lot Mathis!

Issue #650, opened 2022-03-03
Send Every Wallet some Native Asset in Random Wallet test


bfish713:
We send each wallet some Eth but no CAPE native asset, we need both to test all operations

Issue #648, opened 2022-03-03
Wallets Select Only Assets they Can Freeze/Unfreeze Before Trying those Operations in Integration Test


bfish713:
For now in the test the freezer just selects a random asset in the target wallet and tries to freeze it.  This will fail most of the time

Issue #647, opened 2022-03-03
Use Exact Amounts for Burn in the Random Wallet Test


bfish713:
Wallet has a limitation that a burn must be for the exact amount of a record so burns in the test

Issue #645, opened 2022-03-03
Add Freezer Keys at Asset Definition time in Random wallet test


bfish713:
In order for the freeze operations to work we need to actually have some wallets with assets they are allowed to freeze

PR #733, opened 2022-03-11
Improving Wallet Integration Random Tests

bfish713:
- Add all operation types to the not in memory test
- Get exact amount of a record before trying to burn
- Add minting to both (not just in one wallet)
- Select freezable assets when doing freeze and unfreeze operations.

Closes: https://github.com/EspressoSystems/cape/issues/647
Closes: https://github.com/EspressoSystems/cape/issues/650
Closes: https://github.com/EspressoSystems/cape/issues/645
Closes: https://github.com/EspressoSystems/cape/issues/648


wallet/src/bin/random_wallet.rs
```diff
@@ -196,8 +205,9 @@ async fn main() {
             asset.definition
         }
         None => {
+            let policy = AssetPolicy::default().set_freezer_pub_key(freeze_key.clone());
```
jbearer, line 208: You must also audit in order to be able to freeze. Otherwise the wallet won't be able to discover any records that it can freeze. Also note that the policy has to reveal everything: amount, address, and blind.

wallet/src/bin/random_wallet.rs
```diff
@@ -226,75 +236,142 @@ async fn main() {
...
-            Ok(status) => {
-                if !status.succeeded() {
-                    // Transfers are allowed to fail. It can happen, for instance, if we get starved
-                    // out until our transfer becomes too old for the validators. Thus we make this
-                    // a warning, not an error.
-                    event!(Level::WARN, "transfer failed!");
+            OperationType::Freeze => {
+                // Choose a random person who we sent our asset at one point
+                // This may fail if they've transfered that asset
+                if rec_addresses.is_empty() {
```
jbearer, line 336: If you audit, you can actually just use this wallet to list records owned by other wallets that this wallet can freeze (using the `records` iterator with some filtering)

jbearer (changes requested):

wallet/src/testing.rs
```diff
@@ -232,6 +242,27 @@ pub async fn mint_token<'a>(
...
+    freezer: &CapeWallet<'a, CapeBackend<'a, ()>>,
+    frozen: FreezeFlag,
+) -> Vec<RecordInfo> {
+    let pks: HashSet<UserPubKey> = freezer.pub_keys().await.into_iter().collect();
+    let records = freezer.records().await;
+    records
+        .filter(|r| {
+            let ro = &r.ro;
+            // We own the record
+            if pks.contains(&ro.pub_key) {
```
jbearer, line 258: This should be `pks.contains(ro.asset_def.policy_ref().freezer_pub_key())`

wallet/src/testing.rs
```diff
@@ -232,6 +242,27 @@ pub async fn mint_token<'a>(
...
     Ok(my_asset)
 }
+
+/// Return records the freezer has access to freeze or unfreeze but does not own.
+/// Will only return records with freeze_flag the same as the frozen arg.
+pub async fn find_freezable_records<'a>(
+    freezer: &CapeWallet<'a, CapeBackend<'a, ()>>,
+    frozen: FreezeFlag,
+) -> Vec<RecordInfo> {
+    let pks: HashSet<UserPubKey> = freezer.pub_keys().await.into_iter().collect();
```
jbearer, line 252: This should be `let pks: HashSet<FreezerPubKey> = freezer.freezer_pub_keys().await.into_iter().collect();`

wallet/src/testing.rs
bfish713: I may be a bit confused here but the I thought that this function would return records stored in this wallet, which we don't own.  Here I am filtering out records with one of our pub keys because we don't want freeze our own assets (I suppose we could). 

Maybe I should also filter for `ro.asset_def.policy_ref().freezer_pub_key()` in `freezer.freezer_pub_keys()`. But from looking at this code: https://github.com/EspressoSystems/seahorse/blob/main/src/lib.rs#L1443-L1468 I thought the only records the wallet would store which do not contain one of the wallets pub keys are those that are freezable by the wallet.

bfish713: I added the additional filtering for the freezer pub keys in the wallet matching the record

jbearer: Oh, sorry, I swapped the meaning `false` and `true` in my head when I was reading this. It makes sense now. I do think the extra explicit check that we can freeze is good, because the fact that the wallet only stores other user's records if they are freezable is a limitation that we plan to remove, so it will eventually store all records that are viewable, even if not freezable.

jbearer (approved):

PR #749, opened 2022-03-15
Faucet service

nyospe:
Bound to whatever the wallet backend is; this is a very simple service, and currently has no checking beyond "Is this a correctly formed UserPubKey?"

Takes a POST of the user's pub_key, transfers the amount specified in launch parameters from the faucet account to that key's address

### DIFF
--- a/wallet/src/bin/faucet.rs
+++ b/wallet/src/bin/faucet.rs
@@ -1,0 +1,160 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the Configurable Asset Privacy for Ethereum (CAPE) library.
+
+// This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+// This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+// You should have received a copy of the GNU General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+//! # The CAPE Faucet
+//!
+
+extern crate cape_wallet;
+
+use async_std::{
+    sync::{Arc, Mutex},
+    task::{spawn, JoinHandle},
+};
+use cape_wallet::{
+    routes::{init_wallet, Wallet},
+    wallet::CapeWalletError,
+    web::default_storage_path,
+};
+use jf_cap::{
+    keys::{UserKeyPair, UserPubKey},
+    structs::AssetCode,
+};
+use rand_chacha::{rand_core::SeedableRng, ChaChaRng};
+
+use serde::{Deserialize, Serialize};
+use snafu::Snafu;
+use std::path::PathBuf;
+use structopt::StructOpt;
+use tide::StatusCode;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "CAPE Faucet Server",
+    about = "Grants a native asset seed to a provided UserPubKey"
+)]
+pub struct FaucetOptions {
+    /// mnemonic for the faucet wallet
+    #[structopt(long = "mnemonic", env = "MNEMONIC", default_value = "")]
+    pub mnemonic: String,
+
+    /// path to the faucet wallet
+    #[structopt(long = "faucet_wallet_path", env = "FAUCET_WALLET_PATH")]
+    pub faucet_wallet_path: PathBuf,
+
+    /// password on the faucet account keyfile
+    #[structopt(long = "faucet_password", env = "FAUCET_PASSWORD", default_value = "")]
+    pub faucet_password: String,
+
+    /// binding port for the faucet service
+    #[structopt(long = "faucet_port", env = "FAUCET_PORT", default_value = "50079")]
+    pub faucet_port: String,
+
+    /// size of transfer for faucet grant
+    #[structopt(long = "grant_size", env = "FAUCET_GRANT", default_value = "5000")]
+    pub grant_size: u64,
+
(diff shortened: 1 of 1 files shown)


wallet/src/bin/faucet_server.rs
```diff
@@ -0,0 +1,163 @@
...
+use tide::StatusCode;
+
+#[derive(Debug, StructOpt)]
+#[structopt(
+    name = "CAPE Faucet Server",
+    about = "Grants a native asset seed to a provided UserPubKey"
+)]
+pub struct FaucetOptions {
+    /// mnemonic for the faucet wallet
+    #[structopt(long = "mnemonic", env = "MNEMONIC", default_value = "")]
```
sveitser, line 41: I will suggest some more explicit env vars: `CAPE_FAUCET_WALLET_MNEMONIC`?

For the others, could we add the `CAPE_` prefix?

nyospe: Sure, that's reasonable.

wallet/src/bin/faucet_server.rs
```diff
@@ -0,0 +1,163 @@
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
```
sveitser, line 1: Should we add a `faucet` crate in the workspace or is that somehow inconvenient?

wallet/src/bin/faucet_server.rs
```diff
@@ -0,0 +1,163 @@
...
+    about = "Grants a native asset seed to a provided UserPubKey"
+)]
+pub struct FaucetOptions {
+    /// mnemonic for the faucet wallet
+    #[structopt(long = "mnemonic", env = "MNEMONIC", default_value = "")]
+    pub mnemonic: String,
+
+    /// path to the faucet wallet
+    #[structopt(long = "faucet_wallet_path", env = "FAUCET_WALLET_PATH")]
+    pub faucet_wallet_path: PathBuf,
```
sveitser, line 46: I had some issues trying to run the the binary.

I think the first issue is that `write_path` may try to write `last_wallet_path` to a directory that doesn't exist (if one did not run the wallet web server, resp. `create_dir_all` inside the wallet's `init_server` before).

If I do make sure the directory is there I get
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: failed to open wallet: cannot open wallet that does not exist', wallet/src/bin/faucet_server.rs:143:14
```
Is there some initialization code I should run to create the faucet wallet?

wallet/src/bin/faucet_server.rs
```diff
@@ -0,0 +1,163 @@
...
+    faucet_server_error(FaucetError::Transfer {
+        msg: source.to_string(),
+    })
+}
+
+async fn request_fee_assets(
+    mut req: tide::Request<FaucetState>,
+) -> Result<tide::Response, tide::Error> {
+    let bytes = req.body_bytes().await?;
+    let pub_key: UserPubKey = bincode::deserialize(&bytes)?;
```
sveitser, line 108: I see we sometimes use `bincode` and sometimes `TaggedBase64`. I'm wondering what the rationale is for using `bincode` here and in the address book.

sveitser (approved):
LGTM but left some suggestions and questions. The only critical one is that the program terminates immediately. It would be nice to have minimal instructions on how to initialize a wallet and how to run the service as well. I'm ok to merge this first and address the other issues later if necessary.

wallet/src/bin/faucet_server.rs
```diff
@@ -0,0 +1,163 @@
...
+    Ok(spawn(app.listen(address)))
+}
+
+#[async_std::main]
+async fn main() -> Result<(), std::io::Error> {
+    tracing_subscriber::fmt().pretty().init();
+
+    // Initialize the faucet web server.
+    //
+    init_web_server(&FaucetOptions::from_args()).await?;
```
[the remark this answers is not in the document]
nyospe: Oops. I forgot to update that after I added another level of indirection.

wallet/src/bin/faucet_server.rs
nyospe: In general, it's more appropriate to use binary (or Json) for POST operations and Base64 for GET (which is embedded in the URL)

wallet/src/bin/faucet_server.rs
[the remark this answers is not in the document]
nyospe: I want to come back to this later with (correct syntax TBD)
```rust
let handle = init_web_server(&FaucetOptions::from_args()).await?;
let val = while match(poll) = handle.poll() {
    Poll::Pending => continue,
    Poll::Ready(val) => val,
};
```

That way, we could also insert a signal handler to invoke `handle.cancel()` and set a flag on the web service state.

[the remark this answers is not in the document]
nyospe: I want to come back to this later with (correct syntax TBD)
```rust
let handle = init_web_server(&FaucetOptions::from_args()).await?;
let val = while match(poll) = handle.poll() {
    Poll::Pending => continue,
    Poll::Ready(val) => val,
};
```

That way, we could also insert a signal handler to invoke `handle.cancel()` and set a flag on the web service state.

wallet/src/bin/faucet_server.rs
Ancient123: TBH: I am gonna need a more clear explainer of how the (configure/setup/run) loop is supposed to work currently to give any meaningful feedback.

wallet/src/bin/faucet_server.rs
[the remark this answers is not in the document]
nyospe: It's already been fixed.

wallet/src/bin/faucet_server.rs
sveitser: I think we may need to discuss first on what this workflow should look like. I'll add it to the discussion topic for the meeting today.

PR #757, opened 2022-03-17
Fix deadlock in wallet caused by double locking mutex

bfish713:
When I was debugging why the wallet integration test, I noticed the program would hang whenever it chose to burn.  The issue was caused by the call to `is_wrapped_asset()`.  The problem is  `is_wrapped_asset` calls `self.assets()` which locks the state mutex while it already holds the lock.  Locking the same mutex twice causes deadlock.  

Fix is just to swap the order (lock after calling self.assets()).  I confirmed this fix works while working on integration testing and seeing the deadlock disappear with this change.


jbearer (approved):
Oops. Good catch and thanks for fixing

Issue #641, opened 2022-03-03
Wallet Integration Tests Use the Real Address Book


bfish713:
For now we are storing the address mapping in a file on the machine.  It means the test only works from 1 machine and there are race conditions possible.

PR #760, opened 2022-03-17
Make Random Wallet Integration Test work with Address book

bfish713:
Really the only thing for use to use the Address book service from in test was not initializing the logging twice in the in memory test.  Random-wallet.rs already start up and use the address book web service. 

However, this led me to uncover a series of issues in the test or related to the test which I fixed and added more logging to tell what is going on.  

Now we are at a state where the in memory random test will run, and complete all operations until it hits what I think are "real" issues (I.e. issues not caused by a bug in test).  This gets us to the point where the test eventually hits "execution reverted: Root not found". 

Closes: https://github.com/EspressoSystems/cape/issues/641


shenkeyao (approved):
Looks good! Just a typo to fix.

Issue #708, opened 2022-03-09
Health check endpoints for services


sveitser:
All APIs (for the demo: relayer, EQS, address book) should have a health check endpoint that responds with HTTP 200 if the service is okay.

- The simplest version can simply return 200 so that anything monitoring the service can tell it is running. If it's not running we would then spin up another instance.
- If desirable the services can do more extensive checks (for example check if their storage is working). We can consider this out of scope for the demo.

@Ancient123 do you have a preference for 
- HTTP status code and payload to return if service is **healthy**? `200, {"status": "available"}` ?
- HTTP status code and payload to return if service is **unhealthy** `503, {"status": "unavailable"}` ?
- Endoint? `/healthcheck` ?

### TODO
- [ ] Health check endpoint for relayer
- [ ] Health check endpoint for EQS
- [ ] Health check endpoint for address book
- [ ] Health check endpoint for faucet service

nyospe:
This looks like something that could be included in the `net` repo and just added by each service.

pictographer:
Manually checked for Address Book, EQS, relayer, faucet, and wallet web_server.

PR #773, opened 2022-03-18
Add /healthcheck route

pictographer:
#708 

No test cases yet.
Need help running the various executables.
Seeking feedback on the approach.


eqs/src/routes.rs
```diff
@@ -171,6 +172,19 @@ pub async fn get_transaction_by_hash(
     }
 }
 
+/// Return a JSON expression with status 200 indicating the server
+/// is up and running. The JSON expression is simply,
+///    {"status": "available"}
+/// When the server is running but unable to process requests
+/// normally, a response with status 503 and payload {"status":
+/// "unavailable"} should be added.
+pub async fn healthcheck() -> Result<tide::Response, tide::Error> {
```
nyospe, line 181: Looks good for an initial version. Is there going to be support for any other status? EQS has a sort of implicit "catching up" state when it starts up, and it might be worth actually representing that somehow.

nyospe (approved):

Ancient123 (approved):

eqs/src/routes.rs
pictographer: Let's discuss this on Zulip.

PR #775, opened 2022-03-18
Allow for first run load of persistence to be default constructed state

nyospe:


eqs/src/entry.rs
```diff
@@ -19,13 +19,16 @@ use async_std::{
...
         )
     } else {
-        let state_persistence = StatePersistence::load(&opt.store_path(), "eqs").unwrap();
-        let query_result_state =
-            Arc::new(RwLock::new(state_persistence.load_latest_state().unwrap()));
+        let state_persistence = StatePersistence::load(&opt.store_path(), "eth_query").unwrap();
+        let query_result_state = Arc::new(RwLock::new(
+            state_persistence
+                .load_latest_state()
+                .unwrap_or_else(|_| QueryResultState::new(opt.verifier_keys())),
```
sveitser, line 30: @nyospe does this now silently ignore a failure to load the state and start from scratch? I think this might hide some issues with loading the state that we may not want to hide. Would prefer at least emitting a log message.

nyospe: I'll try making it conditional on the type of error. If there's an invalid state, we probably want to panic!, but if the state was simply not found, that indicates a new run.

sveitser (approved):
LGTM

PR #779, opened 2022-03-18
Finish asset-related endpoints

jbearer:
* Include more information in `AssetInfo` (symbol, description, icon, and wrapped ERC-20 code)
* Add `updateasset` endpoint
* Add `importasset`/`exportasset` endpoints


wallet/src/wallet.rs
```diff
@@ -122,6 +123,9 @@ pub trait CapeWalletExt<'a, Backend: CapeWalletBackend<'a> + Sync + 'a> {
         fee: u64,
     ) -> Result<TransactionReceipt<CapeLedger>, CapeWalletError>;
 
+    /// Get the ERC-20 asset code that this asset wraps, if this is a wrapped asset.
+    async fn wrapped_asset(&self, asset: AssetCode) -> Option<Erc20Code>;
```
shenkeyao, line 127: I think `wrapped_erc20` or `wrapped_erc20_code` is a better name since the "wrapped asset" should be the `AssetCode` itself and here we want to get the `Erc20Code` that it's wrapped from.

wallet/src/wallet-api/main.rs
```diff
@@ -1711,6 +1761,228 @@ mod tests {
...
+        // Make sure we didn't export the mint info.
+        assert!(import.mint_info.is_none());
+        // Check that all the information besides the mint info is the same.
+        asset.mint_info = None;
+        assert_eq!(asset, import);
+    }
+
+    #[async_std::test]
+    #[traced_test]
+    async fn test_updateasset() {
```
shenkeyao, line 1820: If it's not too much work, it would be nice to also check that we can't update a verified asset (or is it tested in somewhere else?). This test currently only uses a native asset.

shenkeyao (changes requested):

wallet/src/wallet-api/main.rs
jbearer: I think this is tested pretty thoroughly in Seahorse, so it's okay if this test just focuses on the API specific stuff like route parsing and serialization.

wallet/src/wallet.rs
jbearer: Good point, will do

PR #792, opened 2022-03-21
Small Fix for EQS running in Random Tests

### LINKED ISSUE #548: Update Wallet Integration Test With EQS when it is ready
The random wallet integration test can't work without the EQS.  Wallets won't know when they receive any assets, nor get the memo with the information they need to spend those assets without EQS integration.  As such all but the first "sender" wallet will be stuck waiting for som assets to be able to do anything.

bfish713:
Need to keep the temp dir in scope for the life of the test.  If not we'll end up with the EQS spitting out errors because it's persistence directory got cleaned up.

Closes #548


nyospe (approved):
Oops, yep. This binds it until the NetworkInfo goes out of scope, which is presumably sufficient.

PR #795, opened 2022-03-21
Set chain ID in relayer

jbearer:
With this change, I can run the demo-script, use the wallet API to
create a new wallet and key, use the faucet API to transfer to that
key, and then observe the balance using the wallet API.


sveitser (approved):
LGTM

sveitser:
We could consider handle this in one place so that wallet and relayer could use the same code. I think for the EQS we didn't have the problem because we only do calls.

PR #797, opened 2022-03-21
Add Docker build for the full CAPE demo

jbearer:
Need this for Plaza.

Note sure if I'm doing this right and can't test because `docker build` doesn't seem to work on Mac:
```
> [ 3/16] RUN chmod +x /app/target/release/eqs:
#7 0.196 standard_init_linux.go:228: exec user process caused: exec format error
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -93,6 +93,16 @@ jobs:
             target/release/faucet
             target/release/minimal-relayer
 
+      - uses: actions/upload-artifact@v2
+        with:
+          name: demo-executables
+          path: |
+            target/release/address-book
+            target/release/eqs
+            target/release/faucet
+            target/release/minimal-relayer
+            target/release/wallet-api
+
   docker-wallet:
     runs-on: self-hosted
     needs: build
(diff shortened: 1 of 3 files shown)


Ancient123:
Yeah, the docker build won't function the way we use it on a non-linux machine.

The demo script stuff is ok for now, but in the future we should migrate to a `docker compose` based system for bringing up each of the services in their own container.

jbearer:
Interesting, the 2 containers that were already working failed to build and the one I added succeeded. @Ancient123 do you know what happened there?

sveitser:
I would have suggested to run the `cape-demo` inside the nix shell without docker first. This should work as long as the user can install `nix`. If we want to run the demo inside docker the lets run docker compose as @Ancient123 suggested.

I think the current docker image would not work for the demo because

1. There is no `go-ethereum` inside the docker container.
2. There is no `hardhat` inside the docker container.
3. The cape-demo script uses `hivemind` to start the services (this we could easily change though).

We can also use the container we use on the CI (which has nix) then run `cape-demo` inside the nix shell. This then would not require installing nix on the host.

sveitser:
I think if one doesn't mind waiting a really long time for the demo to start the following also works from withing a checkout of this repo:

```
docker run -v $(pwd):/app -it ghcr.io/espressosystems/nix:main sh -c "cd /app; nix-shell --run cape-demo"
```

PR #802, opened 2022-03-22
Add account status information

jbearer:
Depends on EspressoSystems/seahorse#92

### DIFF
--- a/wallet/src/ui.rs
+++ b/wallet/src/ui.rs
@@ -352,11 +352,12 @@ pub struct Account {
     pub scan_index: Option<EventIndex>,
     /// The ending index of a ledger scan for this account's key.
     ///
-    /// If a ledger scan using this account's key is in progress, `scan_to` is the index of the last
-    /// event in the scan's range of interest. Note that the `scan_to` may be less than
-    /// `scan_index`, since the scan will not complete until it has caught up with the main event
-    /// loop, which may have advanced past `scan_to`.
-    pub scan_to: Option<EventIndex>,
+    /// If a ledger scan using this account's key is in progress, `scan_last_discoverable_event` is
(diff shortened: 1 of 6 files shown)


wallet/src/ui.rs
```diff
@@ -344,6 +345,18 @@ pub struct Account {
...
+    /// If a ledger scan using this account's key is in progress, `scan_index` is the index of the
+    /// next event to be scanned.
+    pub scan_index: Option<EventIndex>,
+    /// The ending index of a ledger scan for this account's key.
+    ///
+    /// If a ledger scan using this account's key is in progress, `scan_to` is the index of the last
+    /// event in the scan's range of interest. Note that the `scan_to` may be less than
+    /// `scan_index`, since the scan will not complete until it has caught up with the main event
+    /// loop, which may have advanced past `scan_to`.
+    pub scan_to: Option<EventIndex>,
```
shenkeyao, line 359: I'd prefer using `scan_status: Option<(EventIndex, EventIndex)>` (as in `AccountInfo`) to make it explicit that there are two kinds of scan indices. `scan_to` sounds like an index that we _will_ scan, so I'd actually expect `scan_to >= scan_index`.

This is just my preference though. It may also suffice to just rename the field(s) rather than combine them into a tuple.

shenkeyao (approved):
Overall looks good! Just a comment about the index fields.

wallet/src/ui.rs
jbearer: I agree, but I think the JSON serialization of `Option<(A, B)>` would be either `null` or `[A, B]`, which is a bit unclear and easy to misuse on the frontend. I will rename the fields though, that's a good point

shenkeyao (approved):

Issue #763, opened 2022-03-17
add test in wallet to make sure these native tokens are transferable


philippecamacho:

shenkeyao:
There's currently a test (`test_send` in `wallet/src/wallet-api/main.rs`) for the `send` endpoint that transfers some native tokens from the faucet address to an unfunded address and verifies the balance.

@tri-joe or @sveitser (sorry not sure who of you brought up this issue), could you confirm whether this is enough for the purpose of this issue?

jbearer:
That test uses a mock faucet. I think we want a test that uses the real faucet service and makes an HTTP request to trigger the transfer

PR #808, opened 2022-03-23
Modify faucet initialization and add a test 

shenkeyao:
- Adds a test for faucet transfer.
- Modifies `create_test_network` and the functions it calls to support faucet setup with a provided key pair.
- Initializes faucet wallet with a provided key pair to make sure the associated mnemonic is consistent.
- Fixes chained server calls.

Closes https://github.com/EspressoSystems/cape/issues/763.


nyospe (approved):
Looks reasonable.

PR #809, opened 2022-03-24
Fix: 'Could not start logging: SetLoggerError(())'

sveitser:
- The errors happens when starting the address book.
- Enable env-filter feature of tracing subscriber. (ping @nyospe )
- Do not pass log level argument: We will use environment variables to
  configure the log level if necessary.
- Remove code that ensure a single logger is running. I don't think it's
  needed anymore. (ping @shenkeyao )

See the following for mention of env-filter feature-flag:

https://docs.rs/tracing-subscriber/0.3.9/tracing_subscriber/filter/struct.EnvFilter.html


nyospe (approved):
LGTM, thanks.

PR #810, opened 2022-03-24
Fix Wallet handling of ERC20 asset lookup

bfish713:
The contract will always return something when you do an asset lookup.  In the case where there is no asset registered in the contract we get a default value of all 0s.  In the wallet we now check for the default code to determine if an asset is wrapped or not.  

Updated the mock to act similarly to the contract.
Added very simple check in the backend tests to make sure the fix works on the real contract


wallet/src/wallet.rs
```diff
@@ -196,6 +201,11 @@ impl<'a, Backend: CapeWalletBackend<'a> + Sync + 'a> CapeWalletExt<'a, Backend>
         let mut state = self.lock().await;
 
         let erc20_code = state.backend().get_wrapped_erc20_code(&cap_asset).await?;
+        if erc20_code == default_erc20_code() {
```
jbearer, line 204: I think we should do this check in `CapeWalletBackend::get_wrapped_erc20_code`, and document in the trait definition that `get_wrapped_erc20_code` returns an error (or possibly `None`?) if the asset is not a wrapped asset.

jbearer (changes requested):

jbearer (approved):

PR #811, opened 2022-03-24
Create a utility to split a single faucet wallet across many wallets

jbearer:


faucet/src/bin/faucet-shower.rs
```diff
@@ -0,0 +1,220 @@
...
+        opt.transfer_size
+    );
+    for (_, mnemonic, key) in &children {
+        println!("{} {}", mnemonic, key);
+    }
+
+    // Do the transfers.
+    for (_, _, key) in &children {
+        match parent
+            .transfer(
```
shenkeyao, line 188: Can this be a single transfer to all the keys? Using a for-loop might be preferred here since we want to have a clearer error message and avoid the size being too big, but all our `transfer` calls in Cape and Seahorse seem to do single-receiver transfers, so I'm wondering if we actually don't want to do multiple-receiver transfer.

shenkeyao (approved):
Looks good. Had a comment about multi-receiver transfers, but it's more like a question rather than a review comment.

faucet/src/bin/faucet-shower.rs
jbearer: This would be really nice, but in CAPE the largest verifying key in the contract has only 3 outputs, so to have more than 3 receiver wallets we would have to break the receivers into chunks of 3 at a time, which I think adds unnecessary complexity.

Issue #777, opened 2022-03-18
Hook up Integration Test with Demo Environment


bfish713:
- [ ] Tests take the url of the each service and uses those for wallet backend
- [ ] Tests uses faucet or faucet wallet to send each wallet some native assets
- [ ] Use contract address from demo

PR #812, opened 2022-03-24
Random Wallet Integration Test Connects to Demo Environment

bfish713:
Test can now take URLs to all the demo services and connect to them.  The sender wallet is funded by the faucet service and then transfers some of its assets to the other wallets in the tests.

Here is the example for how to run with connection to demo services: `cargo run --release --bin random-wallet-in-mem -- 5 --demo-connection=true --eqs-url http://localhost:50010 --address-book-url http://localhost:50000 --relayer-url http://localhost:50020 --faucet-url http://localhost:50030 --contract-address 0xCf7Ed3AccA5a467e9e704C703E8D87F634fB0Fc9`

Closes: https://github.com/EspressoSystems/cape/issues/777


sveitser:
I tried this with the local demo and it worked and it helped me find a problem in the docker demo setup :+1: .

wallet/src/bin/random-wallet-in-mem.rs
```diff
@@ -176,6 +240,47 @@ async fn create_wallet<'a>(
...
+    balances: &mut HashMap<UserAddress, HashMap<AssetCode, u64>>,
+) {
+    if !balances.contains_key(addr) {
+        balances.insert(addr.clone(), HashMap::new());
+    }
+
+    let assets = balances.get_mut(addr).unwrap();
+    let balance = *assets.get(asset).unwrap_or(&0);
+
+    assets.insert(*asset, balance + amount);
```
jbearer, line 256: This looks good. FYI though sometimes the `entry` API can make this kind of thing a bit clearer, like
```
let balance_ref: &mut u64 = balances
  .entry(addr)
  .or_default()
  .entry(asset)
  .or_default();
*balance_ref += amount;
```

wallet/src/bin/random-wallet-in-mem.rs
```diff
@@ -218,12 +323,53 @@ async fn main() {
...
+        let network = get_network_from_args(&args);
+        let mut wallet = connect_to_demo_backend(
+            &network,
+            &universal_param,
+            &mut rng,
+            tmp_dirs.last().unwrap().path(),
+        )
+        .await;
+        let pk = wallet
+            .generate_user_key("sending account".into(), None)
```
jbearer, line 338: `generate_user_key` gets called here and in `connect_to_demo_backend`...is that intentional?

wallet/src/bin/random-wallet-in-mem.rs
```diff
@@ -289,10 +437,11 @@ async fn main() {
                 let address = minter.pub_keys().await[0].address();
                 let asset = mint_token(minter).await.unwrap();
                 event!(Level::INFO, "minted custom asset.  Code: {}", asset.code);
-                balances.get_mut(&address).unwrap().insert(
-                    asset.code,
-                    minter.balance_breakdown(&address, &asset.code).await,
-                );
+                let amount = minter.balance_breakdown(&address, &asset.code).await;
+                balances
```
jbearer, line 441: How come we don't call `add_balance` here?

wallet/src/wallet.rs
```diff
@@ -23,6 +23,11 @@ use std::path::Path;
 
 pub type CapeWalletError = WalletError<CapeLedger>;
 
+pub fn default_erc20_code() -> Erc20Code {
+    let zeros: [u8; 20] = [0; 20];
+    Erc20Code(EthereumAddr(zeros))
+}
```
jbearer, line 29: Where is this used?

bfish713: Oops, this was left over from before I merged in the fix for `is_wrapped_asset` to this branch.

wallet/src/bin/random-wallet-in-mem.rs
bfish713: Nope good catch, looks like the key generated in `connect_to_demo_backend` is not used.

jbearer (approved):

PR #821, opened 2022-03-24
Add tests to ensure CAPE.depositErc20 and CAPE.submitBlock are non reentrant.

### LINKED ISSUE #790: Test correct use of ReentrancyGuard
Create a malicious token that makes a re-entrant call with an empty block. Check revert reason.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/52eeebecda140ebaf4ec8752ed119d8288287fac/contracts/security/ReentrancyGuard.sol#L52

philippecamacho:
Closes #790.

### DIFF
--- a/contracts/contracts/MaliciousToken.sol
+++ b/contracts/contracts/MaliciousToken.sol
@@ -64,6 +64,7 @@ contract MaliciousToken is ERC20 {
             address dummyAddress;
             cape.depositErc20(dummyRo, dummyAddress);
         }
+
         if (_runSubmitBlock) {
             TestCAPE.CapeBlock memory dummyBlock;
             cape.submitCapeBlock(dummyBlock);
--- a/contracts/contracts/MaliciousToken.sol
+++ b/contracts/contracts/MaliciousToken.sol
@@ -44,7 +44,7 @@ contract MaliciousToken is ERC20 {
     /**
      * /// @dev pick the submitBlock function when calling back the CAPE contract
      */
-    function runSubmitBlock() public {
+    function selectSubmitBlockAttack() public {
         _runDeposit = false;
         _runSubmitBlock = true;
     }
--- a/contracts/contracts/MaliciousToken.sol
+++ b/contracts/contracts/MaliciousToken.sol
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+
+// Copyright (c) 2022 Espresso Systems (espressosys.com)
+// This file is part of the Configurable Asset Privacy for Ethereum (CAPE) library.
+
+// This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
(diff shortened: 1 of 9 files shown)


contracts/deploy/10_token.ts
```diff
@@ -7,6 +7,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { deployer } = await getNamedAccounts();
 
   await deploy("SimpleToken", { from: deployer, log: true });
+  await deploy("MaliciousToken", { from: deployer, log: true });
```
sveitser, line 10: Is it useful to deploy the malicious token with hardhat? If yes, would it be better to put it into it's own deployment file so we can still do a deployed with `SimpleToken` but without `MaliciousToken` if needed?

contracts/contracts/MaliciousToken.sol
```diff
@@ -0,0 +1,74 @@
...
+     * /// @dev Sets the address for performing the reentrancy attack.
+     */
+    function setTargetContractAddress(address targetContractAddress) public {
+        _targetContractAddress = targetContractAddress;
+    }
+
+    /**
+     * /// @dev pick the depositErc20 function when calling back the CAPE contract
+     */
+    function runDeposit() public {
```
sveitser, line 39:
```suggestion
    function enableDepositAttack() public {
```
Same suggestion for `runSubmitBlock`.

With the current naming I thought this runs the deposit function when looking at the test.

contracts/rust/src/cape/reentrancy.rs
```diff
@@ -0,0 +1,121 @@
...
+        cape_contract.address(),
+        Arc::new(owner_of_erc20_tokens_client.clone()),
+    );
+
+    let sponsor = owner_of_erc20_tokens_client_address;
+
+    let description = erc20_asset_description(&erc20_code, &EthereumAddr(sponsor.to_fixed_bytes()));
+    let asset_code = AssetCode::new_foreign(&description);
+    let asset_def = AssetDefinition::new(asset_code, AssetPolicy::rand_for_test(rng)).unwrap();
+    let asset_def_sol = asset_def.clone().generic_into::<sol::AssetDefinition>();
```
sveitser, line 44: I think a few variables here aren't so useful: `asset_def_sol`, `amount_u256`,  `cape_contract_address`, `owner_of_erc20_tokens_client_address`, `asset_code`, `call`.

sveitser (approved):
LGTM. Just some comments. Nice test to have :+1:

contracts/deploy/10_token.ts
philippecamacho: Yes, a new script for the Malicious Token has been added in 51cd9d0bc1b1accc21ce6f172e09fae451bee9c4.

contracts/rust/src/cape/reentrancy.rs
philippecamacho: These variables have been removed in faa84b3c812dda1a134ba3c7d15d5ee4e15ba23b.

PR #823, opened 2022-03-25
Use edition "2021" and resolver "2"

sveitser:
Virtual manifests (workspaces w/o [package] sections) require explicitly
setting `resolver = "2"`.

> If you are using a virtual workspace, you will still need to
> explicitly set the resolver field in the [workspace] definition if you
> want to opt-in to the new resolver.

https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,4 +1,5 @@
 [workspace]
+resolver = "2"
 
 members = [
     "address_book",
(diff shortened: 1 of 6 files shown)


nyospe (approved):
Exactly the same as my local changes to test the build!

BTW, I'm not sure if the edition field does anything in the rustfmt.toml file. I couldn't find any indication that it did.

PR #827, opened 2022-03-25
Add (and run) cargo-sort

tri-joe:


flake.nix
```diff
@@ -90,6 +90,7 @@
             stableToolchain
 
             cargo-edit
+            cargo-sort
```
nyospe, line 93: It would be nice to note the invocation (e.g. `cargo sort -w` locally, `cargo sort -w -c` for CI) somewhere. Maybe in README.md?

nyospe (approved):
Good.

PR #833, opened 2022-03-25
Deploy/t194 contract deployment testnet

### LINKED ISSUE #194: Contract deployment on testnet
- [x] #617
- [x]  Run the deployment script
- [x] #839
- [x] Run the smoke tests (defined in #839)
- [x]  Publish the address of deployed contract somewhere (website?)
- [x] Commit the deployment information (json file) contracts/deployments/goerli into the CAPE repository
-  [x] Key generation
    -  [x] Use AZTEC public parameters
    -  [x] Plonk verifying keys
- [x] Define (duplication of #210)
  -  Set height of the merkle tree
  - Set number of roots: 1000 since #513 
- [x] #604
  + If we use ropsten, a repo to setup a miner and mine is at https://github.com/SpectrumXYZ/ethminer-ropsten
 - [x] CAP native token faucet @alxiong 
   - [x] update the public address of the faucet wallet in our deployment script (since we change our mnemonic) #616 
   - [x] #763 
- [x] #834

philippecamacho:
Closes #194.

The CAPE contract has been deployed on Goerli, see https://goerli.etherscan.io/address/0x2006Ac20bA4f826c6561E8fB75ADcde42ae552aA#code

Two questions:
* It seems a good idea to publish the code on etherscan. I tried but it did not work. Ticket #834 created.
* We cannot run the e2e tests because these tests use the mock contract `TestCAPE` instead of `CAPE`. Is that a problem?


contracts/deploy/20_malicious_token.ts
```diff
@@ -1,14 +0,0 @@
-import { HardhatRuntimeEnvironment } from "hardhat/types";
```
sveitser, line 1: @philippecamacho I think if we remove this file we also shouldn't commit `MaliciousToken.json` below.

sveitser:
What is the error you got when trying to do the etherscan validation?

sveitser:
I don't think any of the tests can run unless we set up a faucet correctly. The initial grant is owned by the faucet wallet for which the encryption key / address of the faucet manager are hardcoded in the deployment script. Unless a faucet backend is running with that wallet (or another wallet that has received funds from the faucet manager) the test does not have any native asset balance to pay fees with.

sveitser:
The hardhat deploy plugin also has etherscan verification functionality: https://github.com/wighawag/hardhat-deploy#4-hardhat-etherscan-verify

contracts/deploy/20_malicious_token.ts
philippecamacho: Yes the script has been removed in e63afd8ccbc4d2fc9d7e25397e410ef7350a2e35, but this was after deploying the contract.

philippecamacho:
> I don't think any of the tests can run unless we set up a faucet correctly. The initial grant is owned by the faucet wallet for which the encryption key / address of the faucet manager are hardcoded in the deployment script. Unless a faucet backend is running with that wallet (or another wallet that has received funds from the faucet manager) the test does not have any native asset balance to pay fees with.

Yes, makes sense.Should we add some simple test (some mint and a transfer) that uses this faucet wallet? This would be something one would run  locally (no need to do it on CI) just to be sure the deployment was successful? 
(In all cases I think almost no tests will run anyways because we are using TestCAPE.sol instead of CAPE.sol).

philippecamacho:
@sveitser let us close this PR and create a new one once with have the smoke tests.

PR #840, opened 2022-03-28
Implement an endpoint to build, but not submit, a sponsor transaction

jbearer:


wallet/api/api.toml
```diff
@@ -371,6 +371,58 @@ DOC = """
 Sponsor or define an asset, depending on if an ERC20 code is given. Reports the asset definition.
 """
 
+[route.sponsor]
```
shenkeyao, line 374: Now that we have this endpoint and `newasset` both for sponsor, can we combine them into one `sponsor` endpoint, and add a `submit` or `contract` bool to indicate whether to submit the transaction? Or we could have two endpoints, e.g., `sponsor` and `build_sponsor`. I think either option is clearer than the current structure. (I'll also modify the new `wrap` endpoint to be consistent with your choice.)

Then `newasset` will be specifically for domestic assets, so we can remove sponsor-related paths and rename `newasset` to `create` or `define`, to distinguish it from `sponsor`.

jbearer: I like that. The submit and non-submit cases return different types, so I'll make two endpoints `sponsor` and `buildsponsor`.

jbearer: Actually I'll do `buildsponsor` and `submitsponsor` (which takes the output of `buildsponsor` and the ERC20 code as its only inputs) so that we don't have to duplicate the combinatorial explosion of optional parameters

shenkeyao: Sounds good! I'll add `buildwrap` and `submitwrap` then.

shenkeyao (approved):

PR #858, opened 2022-03-29
Watch directories for contracts build.rs script

### LINKED ISSUE #841: Contracts build.rs script does not watch directories
This is probably why the build.rs script is not re-run after one pulls in changes that add new solidity files.

sveitser:
Close #841

### DIFF
--- a/address_book/Cargo.toml
+++ b/address_book/Cargo.toml
@@ -18,7 +18,7 @@ bincode = "1.3.3"
 dirs = "4.0.0"
 futures = "0.3.21"
 hex = "0.4.3"
-itertools = "0.10.1"
+itertools = "0.10.3"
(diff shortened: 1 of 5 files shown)


philippecamacho (approved):
LGTM

sveitser:
For me on `main`: 

```bash
cargo build 
cargo build -vv # does not re-run the build script
touch contracts/contracts/Foo.sol
cargo build -vv # does not re-run the build script (but should)
```

With this change:
```bash
cargo build 
cargo build -vv # does not re-run the build script
touch contracts/contracts/Foo.sol
cargo build -vv # does re-run the build script 
```
Although the exact behavior of what happens if the contract is created because it was checked out with git may differ.

PR #860, opened 2022-03-29
Ensure deployed faucet record opening is correct

### LINKED ISSUE #787: Add test to check "hardhat deploy" deploys with valid faucet manager keys
I think we can hardcode the encryption key somewhere in the rust codebase and call `hardhat deploy` from a rust test and then verify the data emitted in the event matches what's expected. The `faucet-gen-typescript.rs` should also be changed to use the hardcoded encryption key.

sveitser:
- Hardcode the UserPubKey in the contract crate.
- Run `hardhat deploy` in a rust test and compare faucet record opening.

Close #787

### DIFF
--- a/contracts/rust/src/cape/faucet.rs
+++ b/contracts/rust/src/cape/faucet.rs
@@ -5,90 +5,132 @@
 // This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 // You should have received a copy of the GNU General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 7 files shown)


contracts/rust/src/cape/faucet.rs
```diff
@@ -5,90 +5,132 @@
...
-use anyhow::Result;
-use ethers::abi::AbiDecode;
-use jf_cap::{
-    keys::UserKeyPair,
-    structs::{AssetDefinition, BlindFactor, FreezeFlag, RecordCommitment, RecordOpening},
-    BaseField, MerkleTree,
-};
-
-#[tokio::test]
-async fn test_faucet() -> Result<()> {
```
sveitser, line 25: This test hasn't been changed. It's just indented now.

alxiong (approved):

Issue #831, opened 2022-03-25
wrap endpoint


jbearer:
The `wrap` endpoint we currently have calls the contract using a local ETH wallet. For the GUI MetaMask integration, we should have another variant which does not call the contract, but merely returns the ERC-20 address and a `cap_rust_sandbox::types::RecordOpening`, in a form which the client can easily pass to MetaMask.

PR #861, opened 2022-03-29
Implement buildwrap and submitwrap endpoints

shenkeyao:
Splits the `wrap` endpoint into:
* `buildwrap` which doesn't invoke the contract, for GUI integration
* `submitwrap` which submits the transaction.

Closes https://github.com/EspressoSystems/cape/issues/831.


wallet/src/wallet.rs
```diff
@@ -138,6 +138,27 @@ pub trait CapeWalletExt<'a, Backend: CapeWalletBackend<'a> + Sync + 'a> {
...
+        amount: u64,
+    ) -> Result<RecordOpening, CapeWalletError>;
+
+    /// Submit a wrap transaction to the CAPE contract.
+    ///
+    /// `ro` should be the record opening returned by `build_wrap`.
+    async fn submit_wrap(
+        &mut self,
+        src_addr: EthereumAddr,
+        cap_asset: AssetDefinition,
```
jbearer, line 158: Why do we need the `AssetDefinition`? It should always be the same as `ro.asset_def` right?

jbearer (changes requested):

wallet/src/wallet.rs
shenkeyao: Good catch! In `submitwrap` I actually got the `AssetDefinition` from `ro.asset_def` and passed it to `wallet.submit_wrap`, so yeah, I can just pass `ro` to `wallet.submit_wrap`.

jbearer (approved):

PR #862, opened 2022-03-29
Store named keystores using base64-encoded file names.

### LINKED ISSUE #835: Wallet storage path uses a user input directly as a file name
Instead of building the storage path directly out of the keystore name, we should encode the keystore name in a way that ensures there will be no characters with a special meaning to the file system, such as slashes. URL-safe base64 or TaggedBase64 should both work.

jbearer:
This allows special characters like slashes to be included in
keystore names without being interpreted by the file system.

This change also refactors `lastusedkeystore` to return both the
path and, if applicable, the keystore name, since the name can no
longer be read directly from the file path.

Closes #835

### DIFF
--- a/wallet/src/wallet-api/main.rs
+++ b/wallet/src/wallet-api/main.rs
@@ -88,12 +88,10 @@ mod tests {
     };
     use serde::de::DeserializeOwned;
     use std::collections::hash_map::HashMap;
-    use std::collections::HashSet;
     use std::convert::TryInto;
     use std::fmt::Debug;
     use std::io::Cursor;
     use std::iter::once;
-    use std::iter::FromIterator;
     use std::path::{Path, PathBuf};
     use surf::Url;
     use tempdir::TempDir;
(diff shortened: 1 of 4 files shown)


bfish713 (approved):

wallet/src/wallet-api/main.rs
```diff
@@ -1779,38 +1779,17 @@ mod tests {
...
-        let expected: HashSet<String> =
-            vec![String::from("named_keystore"), String::from("test_wallet")]
-                .into_iter()
-                .collect();
-
-        assert_eq!(
-            expected,
-            HashSet::from_iter(from_server_vec.iter().cloned())
-        );
-
```
bfish713, line 1811: Does this no longer work? Or does it just not really make sense as a test anymore?

jbearer: It doesn't work (because when listing we skip files that aren't valid tagged base 64) and it doesn't make sense anymore (because this behavior isn't part of the interface/spec of `newwallet`/`listkeystores`, and never was, really)

PR #877, opened 2022-03-30
Misc doc cleanup

sveitser:
- Add README.md to all crates. Include them as crate doc where it makes
  sense. A problem here is that the markdown links don't work.
- Make sure all crates (except random wallet test) have at least a one
  sentence summary in cargo doc.
- Add some missing copyright headers.
- [x] disable docs for binaries

### DIFF
--- a/faucet/src/faucet_wallet_test_setup.rs
+++ b/faucet/src/faucet_wallet_test_setup.rs
@@ -5,6 +5,9 @@
 // This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 // You should have received a copy of the GNU General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! A script to export enviornment variables for test deployments of the CAPE
+//! contract with hardhat.
+
 use cap_rust_sandbox::helpers::compute_faucet_key_pair_from_mnemonic;
 use cap_rust_sandbox::types::EdOnBN254Point;
 use ethers::{abi::AbiEncode, prelude::U256};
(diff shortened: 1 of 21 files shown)


faucet/src/faucet_wallet_test_setup.rs
```diff
@@ -5,6 +5,9 @@
 // This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 // You should have received a copy of the GNU General Public License along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+//! A script to export enviornment variables for test deployments of the CAPE
```
pictographer, line 8: environment

pictographer (approved):
Aside from one typo, looks very good to me.

philippecamacho:
Is it normal to still have these warnings?
```
[nix-shell:~/repositories/Translucence/cape]$ cargo doc --open
warning: output filename collision.
The lib target `atomic_store` in package `atomic_store v0.1.0 (https://github.com/EspressoSystems/atomicstore.git#b036409b)` has the same output filename as the lib target `atomic_store` in package `atomic_store v0.1.0 (https://github.com/EspressoSystems/atomicstore.git?tag=0.1.0#a719ad0e)`.
Colliding filename is: /home/philippe/repositories/Translucence/cape/target/doc/atomic_store/index.html
The targets should have unique names.
This is a known bug where multiple crates with the same name use
the same path; see <https://github.com/rust-lang/cargo/issues/6313>.
```

philippecamacho (approved):
Added a single question. LGTM

sveitser:
> Is it normal to still have these warnings?

@philippecamacho Yes. This seems to go away if we specify the `atomic_store` dependency in the same way in every `Cargo.toml` in the workspace. We have similar warnings for ethers but there I think it comes from ethers using a different way to specify the version to refer to internal ethers dependencies. If we don't specify `branch = "master"` the warnings go away.

I made this change in 33685d983d43f900e9ac5b002db896e032501fd0 . @nyospe is it ok to track the latest `atomic_store` for all the crates in this workspace?



Btw, to only build the documentation for this workspace we can run
```
cargo doc --open --no-deps
```

sveitser:
Converted to draft again. @nyospe will check if it's okay to not pin `atomic_store` to `tag=0.1.0`.

sveitser:
After #911 is merged only the EQS will depend on atomic store and the above should be gone.

Issue #875, opened 2022-03-30
Random wallet test panics


sveitser:
I was running the random wallet test against the local demo and got a panic after running about 2 minutes at

https://github.com/EspressoSystems/cape/blob/d932ddb2ab758205c730ecce6b79e06050e6b6e0/wallet/src/bin/random-wallet-in-mem.rs#L458

```
  2022-03-30T09:51:00.366617Z  INFO random_wallet_in_mem: Unfreezing
    at wallet/src/bin/random-wallet-in-mem.rs:525

  2022-03-30T09:51:00.366647Z  INFO random_wallet_in_mem: No frozen records
    at wallet/src/bin/random-wallet-in-mem.rs:531

  2022-03-30T09:51:00.366653Z  INFO random_wallet_in_mem: Transfering
    at wallet/src/bin/random-wallet-in-mem.rs:436

thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', wallet/src/bin/random-wallet-in-mem.rs:458:61
stack backtrace:
   0: rust_begin_unwind
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:107:14
   2: core::panicking::panic
             at /rustc/db9d1b20bba1968c1ec1fc49616d4742c1725b4b/library/core/src/panicking.rs:48:5
   3: random_wallet_in_mem::main::main::{{closure}}
   4: std::thread::local::LocalKey<T>::with
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   6: async_io::driver::block_on
   7: std::thread::local::LocalKey<T>::with
   8: async_std::task::builder::Builder::blocking
   9: random_wallet_in_mem::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```
Full log:
[random-wallet.log](https://github.com/EspressoSystems/cape/files/8379516/random-wallet.log)

cc @bfish713

bfish713:
[...]

PR #878, opened 2022-03-30
Fix panic where wallet has no assets to transfer

bfish713:
Closes: https://github.com/EspressoSystems/cape/issues/875


sveitser (approved):
LGTM

PR #890, opened 2022-03-31
Certificates not found in docker container

sveitser:
Create /etc/ssl/certs/ca-certificates.crt

cc @DieracDelta @nmccarty in case you have an idea why `/etc/ssl/certs/ca-certificates.crt` would be missing in the docker container (even if `cacert` added).

edit; Why it's there on nixos: https://github.com/NixOS/nixpkgs/blob/9a813114b916d3abffd60b86a91ea707d98011cd/nixos/modules/security/ca.nix#L76

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,7 +26,22 @@
     }:
     flake-utils.lib.eachDefaultSystem (system:
     let
-      overlays = [ (import rust-overlay) ];
+      overlays = [
+        (import rust-overlay)
+        (final: prev:
+          let
+            runInLinuxVMNoKVM = drv:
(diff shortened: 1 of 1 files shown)


Ancient123 (approved):

DieracDelta:
I'm not sure why this isn't symlinked by default. It looks like [`docker-nixpkgs`](https://github.com/nix-community/docker-nixpkgs/blob/master/images/nix/default.nix#L59) avoids the issue with an environment variable.

sveitser:
Maybe setting the env var like that won't break the github docker build. Will try first thing tomorrow.

Ancient123:
We can try that tomorrow as well. For now this *should* unblock me for some attempts tonight.

PR #904, opened 2022-04-01
Remove tide-tracing + bypass cargo-clippy bug

nyospe:
tide-tracing had hardcoded `info` level messages on request entry/exit, with no mechanism for changing them to `trace` level. I'll add a tide middleware processor for tracing to the `net` repo instead.


sveitser (approved):
LGTM

philippecamacho:
@nyospe why is the address book still using tide for logging?
https://github.com/EspressoSystems/cape/blob/bb853d5c03536a80517a734602f8962bbdb998e5/address_book/src/main.rs#L32

PR #905, opened 2022-04-01
A less involved fix for SSL certs

sveitser:
Works for `curl` but not yet tested for our rust service.

Edit: worked for starting up the EQS and connecting to geth on a https domain.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,22 +26,7 @@
     }:
     flake-utils.lib.eachDefaultSystem (system:
     let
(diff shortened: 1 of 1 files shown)


DieracDelta:
LGTM this is definitely cleaner. I'm surprised this fixed the `kvm` problem.

DieracDelta (approved):

sveitser:
I think the kvm issue only showed up after adding the `runAsRoot` step.

PR #907, opened 2022-04-04
Review of logging for the CAPE services.

philippecamacho:
I tried to log the services options for the eqs, relayer and faucet. However it seems to only work for the faucet.

See discussion https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/tracing.3A.3Ainfo!.20.20does.20not.20seem.20to.20work.20for.20the.20relayer.2Feqs/near/277752712


relayer/src/api_server.rs
```diff
@@ -52,11 +51,11 @@ pub(crate) fn init_web_server(
     txn_queue: Arc<RwLock<TxnQueue>>,
 ) -> Result<task::JoinHandle<Result<(), std::io::Error>>, tide::Error> {
     let mut web_server = tide::with_state(WebState { txn_queue });
-    web_server.with(TraceMiddleware::new());
     web_server.at("/healthcheck").get(healthcheck);
     web_server.at("/submit").post(submit_endpoint);
     let port = std::env::var("PORT").unwrap_or_else(|_| DEFAULT_RELAYER_PORT.to_string());
     let addr = format!("0.0.0.0:{}", port);
+    tracing::info!("RELAYER ADDRESS {}", addr);
```
nyospe, line 58: This is not the `init_web_server` called by the minimal-relayer service.

philippecamacho: Thanks @nyospe , I will close the PR then.

PR #910, opened 2022-04-05
Add script to generate Ethereum mnemonic

sveitser:
Running `generate-eth-mnemonic` inside the nix shell will print a random mnemonic and its first address.


philippecamacho (approved):
I tried to use the mnemonic in metamask and it worked.

LGTM

PR #911, opened 2022-04-05
Remove dead code of the Relayer

### LINKED ISSUE #909: Remove dead code of the full relayer
See https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/.E2.9C.94.20tracing.3A.3Ainfo!.20.20does.20not.20seem.20to.20work.20for.20the.20relayer.2Feqs/near/277761129

philippecamacho:
Closes #909.


relayer/src/main.rs
```diff
@@ -1,66 +0,0 @@
...
-mod configuration;
-mod state_persistence;
-mod txn_queue;
-
-#[async_std::main]
-async fn main() -> std::io::Result<()> {
-    tracing_subscriber::fmt()
-        .compact()
-        .with_ansi(false)
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
```
sveitser, line 34: I checked unused dependencies with cargo udeps and found

```
atomic_store
dirs
tide-websockets
toml
tracing-log
tracing-subscribers
```
The code still compiles without them.

However for the tracing it's because it was set up in`main.rs` and is therefore missing now. We should move this to the minimal relayer binary. But also remove the truly unused dependencies.

sveitser (changes requested):
- [x] Move logging setup to minimal relayer.
- [ ] Remove unused depedencies.

relayer/src/main.rs
philippecamacho: Thank you for spotting that.
Changes done in d3ede09.

sveitser (approved):
LGTM

PR #912, opened 2022-04-06
EQS: set the number of Eth block confirmations

sveitser:
- Add `Confirmations` newtype.
- Add `CAPE_EQS_NUM_CONFIRMATIONS` env var.
- Panic if restoring state tries to increase the number of
  confirmations. This is not possible with the current design.

Should still add some tests, set the number of confirmations to more
than 1 for some tests and/or test this manually.

The `cape-demo-local` script is changed from instant mining to using
a 1 second block time. To get the instant mining behavior back one can
run `env GETH_PERIOD=0 cape-demo-local`. This won't work well
for `CAPE_EQS_NUM_CONFIRMATIONS > 1` because transactions won't
be confirmed unless new transactions are sent to geth to generate new
blocks.


eqs/src/configuration.rs
```diff
@@ -7,9 +7,44 @@
...
+        Self(1)
+    }
+}
+
+impl FromStr for Confirmations {
+    type Err = ParseIntError;
+    fn from_str(s: &str) -> Result<Self, ParseIntError> {
+        let n = s.parse::<u64>()?;
+        if n == 0 {
+            panic!("Confirmations must be >= 1")
```
nyospe, line 41: StructOpt produces better exit messages if you return the error, rather than invoking the panic!

nyospe (approved):
👍

eqs/src/configuration.rs
sveitser: I don't know how to create a `ParseIntError`. Is it impossible? https://stackoverflow.com/a/55616075

Or do you mean defining my own error type for this instead?

nyospe: I meant define an error type to emit instead. But there's actually an exiting type you could use instead.

https://doc.rust-lang.org/std/num/struct.NonZeroU32.html#impl-FromStr

Somehow, this still ends up producing a ParseIntError.

sveitser:
EQS cannot be started
```
eqs          | thread 'main' panicked at 'Cannot increase the number of confirmations: persisted state uses Confirmations(1) which is smaller than Confirmations(6)', /Users/lulu/r/EspressoSystems/cape/eqs/src/eth_polling.rs:72:17
```
I think it's because the store is first initialized with defaults `Confirmations(1)`.

PR #922, opened 2022-04-07
Configure wallet Docker to connect to geth deployment

jbearer:


sveitser (approved):
LGTM

sveitser:
@jbearer @Ancient123 I wonder if we just want to make goerli the default?

jbearer:
@sveitser eventually yes (like probably within a few days) but I want to get this to Plaza today and we need to figure out the web3 query service issue for Goerli

Issue #892, opened 2022-03-31
Return asset metadata with account balances


jbearer:
Currently, the UI must make two separate requests for the accounts page: one to get balances, and one to get the metadata for the corresponding assets. We can improve performance by returning the asset metadata with the balances.

Note that this is not required for release but would be a UX improvement,

philippecamacho:
Optional, would make the GUI more efficient.

PR #925, opened 2022-04-08
Add Map from asset code to asset info when returning balances

bfish713:
Return a map of the known assets along with the balances.  This will make the UI a bit more efficient by not having to call `getinfo` to get the asset info after it calls `getbalance`

Cloeses: https://github.com/EspressoSystems/cape/issues/892

### DIFF
--- a/wallet/src/ui.rs
+++ b/wallet/src/ui.rs
@@ -313,14 +313,19 @@ pub enum PrivateKey {
     Freezing(FreezerKeyPair),
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum BalanceInfo {
     /// The balance of a single asset, in a single account.
     Balance(u64),
     /// All the balances of an account, by asset type.
-    AccountBalances(HashMap<AssetCode, u64>),
+    AccountBalances((HashMap<AssetCode, u64>, HashMap<AssetCode, AssetInfo>)),
     /// All the balances of all accounts owned by the wallet.
-    AllBalances(HashMap<UserAddress, HashMap<AssetCode, u64>>),
+    AllBalances(
+        (
+            HashMap<UserAddress, HashMap<AssetCode, u64>>,
+            HashMap<AssetCode, AssetInfo>,
+        ),
+    ),
 }
 
 #[ser_test(ark(false))]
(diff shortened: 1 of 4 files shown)


wallet/src/ui.rs
```diff
@@ -313,14 +313,19 @@ pub enum PrivateKey {
...
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum BalanceInfo {
     /// The balance of a single asset, in a single account.
     Balance(u64),
     /// All the balances of an account, by asset type.
-    AccountBalances(HashMap<AssetCode, u64>),
+    AccountBalances((HashMap<AssetCode, u64>, HashMap<AssetCode, AssetInfo>)),
```
shenkeyao, line 321: Since each `AssetCode` corresponds to a balance and an `AssetInfo`, I think it's better to have just one map, i.e., `HashMap<AssetCode, (u64, AssetInfo)>` if this doesn't make the enum too complicated to construct.

wallet/src/ui.rs
```diff
@@ -313,14 +313,19 @@ pub enum PrivateKey {
     Freezing(FreezerKeyPair),
 }
 
-#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
 pub enum BalanceInfo {
     /// The balance of a single asset, in a single account.
     Balance(u64),
     /// All the balances of an account, by asset type.
```
shenkeyao, line 320: The comment needs to be updated.

wallet/src/ui.rs
```diff
@@ -313,14 +313,19 @@ pub enum PrivateKey {
...
-    AccountBalances(HashMap<AssetCode, u64>),
+    AccountBalances((HashMap<AssetCode, u64>, HashMap<AssetCode, AssetInfo>)),
     /// All the balances of all accounts owned by the wallet.
```
shenkeyao, line 322: Same here, we need to update the comment.

shenkeyao (changes requested):

wallet/src/ui.rs
bfish713: I did it this way to avoid sending the same AssetInfo struct multiple times when you get all the balances, but that's probably a non issue.  I can make this change I don't think it's any more complicated now

PR #933, opened 2022-04-09
Use request body for updateasset parameters to avoid URL-encoding icon

jbearer:


bfish713 (approved):
Looks good, thanks for jumping on this

PR #942, opened 2022-04-12
Use ubuntu as base to build docker

Ancient123:


sveitser (approved):
LGTM. The script to build the wallet docker image locally will need to be updated but I currently can't test this easily because I only have my mac with me.

Created 
- https://github.com/EspressoSystems/cape/issues/944

to track that.

PR #948, opened 2022-04-13
Updated /healthcheck to respond with Initializing or Available

pictographer:
Previously, the `/healthcheck` endpoint would always respond
```json
  {"status": "available"}
```
Now, until it finishes catching up, it responds
```json
  {"status": "Initializing"}
```
and 
```json
  {"status": "Available"}
```
Note the status value is capitalized to match the Rust enumeration naming convention. This can be fixed if it's a problem.

Thanks to a suggestion from Mathis, the system status is stored as an enumeration, so this will be easy to expand.

- [ ] Find a way to test this.

### DIFF
--- a/eqs/src/query_result_state.rs
+++ b/eqs/src/query_result_state.rs
@@ -44,6 +44,8 @@ pub struct QueryResultState {
     // additional indexed data for queries
     pub transaction_by_id: HashMap<(u64, u64), CommittedCapeTransition>,
     pub transaction_id_by_hash: HashMap<Commitment<CapeTransition>, (u64, u64)>,
+
+    pub catching_up: bool,
 }
 
 impl QueryResultState {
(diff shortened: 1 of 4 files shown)


eqs/src/query_result_state.rs
```diff
@@ -44,6 +51,8 @@ pub struct QueryResultState {
     // additional indexed data for queries
     pub transaction_by_id: HashMap<(u64, u64), CommittedCapeTransition>,
     pub transaction_id_by_hash: HashMap<Commitment<CapeTransition>, (u64, u64)>,
+
+    pub system_status: SystemStatus,
```
nyospe, line 55: I wonder if this should actually be set back to Initializing on Deserialize...

nyospe: Should be possible with `#[serde(skip)]` and `#[serde(default = "assign_initializing")]` with an `fn assign_initializing() -> SystemStatus { SystemStatus::Initializing }`

pictographer: Oh, good catch! Loading system_status of Available from persistent store doesn't make sense if there might be an unbounded amount of work to do after deserialization.

Ancient123 (approved):

PR #950, opened 2022-04-14
Fix flaky hardhat deploy test

sveitser:
This will run the test in isolation and prevent it from failing.

In case it also fails stderr from running `hardhat deploy` will 
now show up in the CI logs.

Also combines the other integration tests in a single binary. This
should reduce compilation and test time. See the "Modularizing tests"
section of https://mudit.blog/rust-tips-and-tricks/.


philippecamacho (approved):
LGTM

PR #955, opened 2022-04-14
Do not add sponsored asset to library until we know it is ok.

jbearer:
Previously, the buildsponsor endpoint (and the corresponding
`build_sponsor` wallet method) would add the new AssetInfo to the
wallet's library before returning the `sol::AssetDefinition` to be
submitted to the contract. This can lead to confusing UX if the user
cancels the transaction before the asset is actually sponsored, or
if the Ethereum transaction is rejected or reverted, because it
would lead to an asset in the asset library which doesn't actually
correspond to anything.

Now, `buildsponsor` just returns the serialized AssetInfo instead of
adding it to the wallet. The asset can be added to the wallet at any
time after the sponsor transaction succeeds by calling `importasset`.

### DIFF
--- a/wallet/src/wallet.rs
+++ b/wallet/src/wallet.rs
@@ -98,7 +98,6 @@ pub trait CapeWalletExt<'a, Backend: CapeWalletBackend<'a> + Sync + 'a> {
     /// does, use [CapeWalletExt::sponsor].
     async fn build_sponsor(
         &mut self,
-        symbol: String,
         erc20_code: Erc20Code,
         sponsor_addr: EthereumAddr,
         cap_asset_policy: AssetPolicy,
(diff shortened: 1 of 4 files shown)


wallet/src/wallet.rs
```diff
@@ -98,7 +98,6 @@ pub trait CapeWalletExt<'a, Backend: CapeWalletBackend<'a> + Sync + 'a> {
     /// does, use [CapeWalletExt::sponsor].
     async fn build_sponsor(
```
shenkeyao, line 99: I think we should remove "added to the asset library" from the comment.

shenkeyao (changes requested):
Looks good! Just need to update the comment for `build_sponsor`.

wallet/src/wallet.rs
jbearer: Oops good catch

shenkeyao (approved):

Issue #952, opened 2022-04-14
Relayer should log the hash of the ethereum transaction before submitting a CAP block


philippecamacho:
When the wallet submits a CAP tx and the corresponding ethereum transaction does not go through the wallet has no way of knowing it can respend the outputs. See Zulip [discussion](https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/Error/near/278767118).

In order to make it easier debugging this kind of issues the relayer should log the hash of the ethereum transaction before submitting a CAP block.

pictographer:
https://github.com/EspressoSystems/cape/pull/959

pictographer:
Closing this because I've created another pull request that is up to date and doesn't need commits pruned out of it. https://github.com/EspressoSystems/cape/pull/973

PR #959, opened 2022-04-15
Relayer log message with ETH transaction hash

pictographer:
Relayer should log the hash of the ethereum transaction before submitting a CAP block [#952](https://github.com/EspressoSystems/cape/issues/952)


relayer/src/lib.rs
```diff
@@ -173,6 +174,7 @@ async fn relay(
     // The pending transaction itself doesn't serialize well, but all the relevant information is
     // contained in the transaction hash. The client can reconstruct the pending transaction from
     // the hash using a particular provider.
+    event!(Level::INFO, "ETH H256: {}", *pending);
```
sveitser, line 177:
```suggestion
    event!(Level::INFO, "Submitted ethereum transaction hash: {:?}", *pending);
```
The exact wording is not so important but I think searching for "transaction hash" should find these logs.

Also without the debug formatting it prints an abbreviated form:

```
0x8ca3…9922
```
instead of
```
0x3eac86d83469d6c1f95d4776b5fe276df1321c761e64311d534feef09339eb45
```
The abbreviated form would make it difficult or impossible to lookup in a blockchain explorer.

sveitser (changes requested):

relayer/src/lib.rs
jbearer: FYI you can also use `{:x}` or `{:#x}` to print the full hex address without relying on `Debug`

sveitser:
Superseded by #973.

PR #963, opened 2022-04-18
Add shell for static linking

DieracDelta:
Initial attempt at a nix shell that provides tooling for building statically linked rust binaries for the workspace. Usage: `nix develop .#staticShell -c cargo build --release`. Note that I have not tested this beyond checking that all binary targets compile, `wallet-api` runs, and `ldd wallet-api` comes back empty. I haven't set up a linux box with FDE to test on, and `nix-portable` fails to initialize the shell.

EDIT: what is used to format the `flake.nix`? I used `nixpkgs-fmt` and it looks like it changed a bunch of stuff.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,117 +33,69 @@
     }:
     flake-utils.lib.eachDefaultSystem (system:
     let
-      overlays = [ (import rust-overlay) ];
-      pkgs = import nixpkgs {
-        inherit system overlays;
+      xpkgs = import nixpkgs {
+        system = "x86_64-linux"; 
+        overlays = [ (import rust-overlay) ];
       };
+      pkgs = import nixpkgs {
+        localSystem = "x86_64-linux"; 
+        crossSystem = { config = "x86_64-unknown-linux-musl"; };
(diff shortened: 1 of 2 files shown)


sveitser:
@DieracDelta yes, `nixpkgs-fmt`. You can also run `treefmt` to format everything in the repo.

sveitser (approved):
Great! LGTM but I think we should mention the feature in the README to make it easier to discover.

flake.nix
```diff
@@ -213,12 +261,18 @@
               export PATH=''${my_pwd}/bin:$PATH
               export WALLET=''${my_pwd}/wallet
 
-              git config --local blame.ignoreRevsFile .git-blame-ignore-revs
-            ''
-            # install pre-commit hooks
-            + self.checks.${system}.pre-commit-check.shellHook;
+              export RUSTFLAGS='-C target-feature=+crt-static'
```
sveitser, line 264: These could also go outside the `shellHook` like some of the other env vars.

Ancient123 (approved):

PR #968, opened 2022-04-19
Check ERC20 address matches registered address

### LINKED ISSUE #954: Deposit function does not check if ERC20 address matches that of registered asset definition
[Zulip thread](https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/depositErc20.20address)

The explicit check (1.) is absent in the contract:

https://github.com/EspressoSystems/cape/blob/16e7fa0573e2a95ff9d089e1d7895d595775cbb6/doc/workflow/src/lib.rs#L236-L244

sveitser:
- Add test for deposit with wrong ERC20 address
- Some refactoring to make the test code more concise.

Close #954

### DIFF
--- a/contracts/contracts/CAPE.sol
+++ b/contracts/contracts/CAPE.sol
@@ -213,6 +213,7 @@ contract CAPE is RecordsMerkleTree, RootStore, AssetRegistry, ReentrancyGuard {
     /// @param erc20Address address of the ERC-20 token corresponding to the deposit
     function depositErc20(RecordOpening memory ro, address erc20Address) public nonReentrant {
         require(isCapeAssetRegistered(ro.assetDef), "Asset definition not registered");
+        require(lookup(ro.assetDef) == erc20Address, "Wrong ERC20 address");
 
         // We skip the sanity checks mentioned in the rust specification as they are optional.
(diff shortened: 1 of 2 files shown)


jbearer (approved):

contracts/contracts/CAPE.sol
```diff
@@ -213,6 +213,7 @@ contract CAPE is RecordsMerkleTree, RootStore, AssetRegistry, ReentrancyGuard {
     /// @param erc20Address address of the ERC-20 token corresponding to the deposit
     function depositErc20(RecordOpening memory ro, address erc20Address) public nonReentrant {
```
alxiong, line 214: Shall we remove the `address erc20Address` in this PR?

philippecamacho: We prefer not to for now to avoid breaking the ui code. @sveitser will create another ticket to do this later.

philippecamacho (approved):
LGTM

contracts/contracts/CAPE.sol
sveitser: The ticket is #971

Issue #949, opened 2022-04-14
EQS panics if HTTP requests to provider fail


sveitser:
Requests could fail for a variety of reasons so I think it would be desirable to return an Error and retry instead of panicking right away.

https://github.com/EspressoSystems/cape/blob/4a0de582065b27df2592fe210576e67639f4e220/eqs/src/eth_polling.rs#L129

https://github.com/EspressoSystems/cape/blob/4a0de582065b27df2592fe210576e67639f4e220/eqs/src/eth_polling.rs#L152

The second request could also hit infura limits of 10k events or over 10 seconds query time: https://docs.infura.io/infura/networks/ethereum/json-rpc-methods/eth_getlogs#limitations

philippecamacho:
Potential solution: change the EQS so that it fetches not all the blocks at once.
If the EQS needs to fetch more than 10k events it will not be able to start at all.

Ancient123:
Related to this, they should absolutely log `ERROR` on request failure. It is definitely a signal we want to watch.

PR #974, opened 2022-04-19
Removing the primary failure-loop scenario, starting to remove unwraps

nyospe:


nyospe:
Addresses the specific case that https://github.com/EspressoSystems/cape/issues/949 refers to.

sveitser:
I'm wondering if we should re-organize the code a bit. I think it would make it easier to review this and future changes to the code.

Could we move the fetch loop (or new inner loop) into it's own method, compute the batches (pairs of `from` and `to` block numbers) and then pass that to the `fetch` method? Is there any drawback to that? What do you think @nyospe?

eqs/src/eth_polling.rs
```diff
@@ -114,44 +116,76 @@ impl EthPolling {
...
+                Some(n) => n,
+                // There is no confirmed block to fetch yet.
+                None => return Ok(0),
+            };
+
+            if fetch_latest < fetch_from {
+                break;
+            }
+
+            let fetch_until = if fetch_latest > fetch_from + self.max_blocks_to_query {
```
sveitser, line 150: @nyospe Is this `min(fetch_latest, fetch_from + self.max_blocks_to_query)`?

sveitser (approved):
LGTM

eqs/src/eth_polling.rs
nyospe: I tried using min initially, but something went wrong. I don't remember what, but I must've changed whatever caused it, because it seems to work now.

nyospe:
Closes https://github.com/EspressoSystems/cape/issues/949

PR #977, opened 2022-04-19
Update dependencies, fixing key-set bug

jbearer:


bfish713:
Nice thanks for doing this, I will have a PR to update seahorse dependency shortly.

PR #982, opened 2022-04-20
Relayer: submit an empty block every 5 minutes

### LINKED ISSUE #958: Relayer sends an empty block periodically
[Post](https://discord.com/channels/857206489433112586/916010667117518859/964234879103537183) from @jbearer  on discord:

> When the user wraps some ERC20, all that does is add a pending wrap action to a queue in the contract. The wrap will not actually be finalized (and thus the balance will not show up in the wallet) until some other transaction (mint, transfer, etc.) goes through. This is so that the relayer, and not the user, pays the majority of the gas for processing wraps.
> 
> @clu @min , do we want to handle this in the UI at all? E.g. we could have a popup message alerting the user that it could be some time before they see the results.
> 
> @thedulwhichfox @sveitser there are things we could do on the backend that would make this better. For example, having the relayer submit an empty block every short time interval (minute, maybe?) would mean that you don't have to wait for someone else to submit a transaction, or submit one yourself, just to finalize your pending wrap. (In addition, it would keep the pending deposit queue from growing.)

sveitser:
The interval for empty block submissions can be configured with the
`CAPE_RELAYER_EMPTY_BLOCK_INTERVAL_SECS` env var.

- Add missing `.init` for tracing subscriber.

Close #958

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4861,6 +4861,7 @@ dependencies = [
  "bincode",
  "cap-rust-sandbox",
  "ethers",
+ "futures",
  "jf-cap",
  "jf-primitives",
  "key-set",
--- a/relayer/Cargo.toml
+++ b/relayer/Cargo.toml
@@ -23,6 +23,7 @@ async-std = { version = "1.10.0", features = ["unstable", "attributes", "tokio1"
 bincode = "1.3.3"
 cap-rust-sandbox = { path = "../contracts/rust" }
 ethers = { git = "https://github.com/gakonst/ethers-rs" }
+futures = "0.3.21"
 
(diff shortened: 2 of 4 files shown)


jbearer (commented):
This looks good but are we not doing the thing where we check if the pending deposits queue is non-empty before submitting the empty block?

sveitser:
We can but I'm not sure it's necessary. It's quite cheap to submit an empty block, it's about 50k gas compared to about 3M gas for processing a single deposit. If we do care about gas we may want to wait until the queue is full or almost full because the merkle tree insertion cost grows sub-linearly with the number of deposits.

jbearer (approved):

PR #987, opened 2022-04-21
Fix geth docker image url

sveitser:


VictorKoenders (approved):
Pulled successfully locally 👍

PR #999, opened 2022-04-28
Add WETH and DAI wrappable tokens for testing

### LINKED ISSUE #996: Add more tokens to local demo backend
There should be a few tokens. We can use `WETH` to make it easier to mint them.

sveitser:
Close #996.

This will mint 1 million tokens for each Ether sent to the token contract.

Haven't tested the unwrapping yet, but we don't really need it for testing unless we care about recovering the Ether.


philippecamacho (approved):
LGTM

PR #1005, opened 2022-05-04
Run an ARM64 build

Ancient123:


Ancient123:
@sveitser : Is there a major reason to use solc-bin and not the emscripten/JS version? I don't see solidity compile times being a major problem, but I didn't know if there was another reason for going this route.

Ancient123:
Alternatively we could be building/installing solc from source, I don't know how time expensive that is in Nix.

Ancient123:
```
.../node_modules/bufferutil install:   CC(target) Release/obj.target/bufferutil/src/bufferutil.o
.../node_modules/bufferutil install: ../src/bufferutil.c:3:10: fatal error: node_api.h: No such file or directory
```

Still seeing something weird at nix/build time. :(

.github/workflows/build.yml
```diff
@@ -1,15 +1,15 @@
...
-  push:
-    branches:
-      - main
-    tags:
-      # YYYYMMDD
-      - "20[0-9][0-9][0-1][0-9][0-3][0-9]"
-  pull_request:
-  schedule:
-    - cron: "0 1 * * 1"
+  #push:
```
sveitser, line 4: @Ancient123 wouldn't this prevent our main CI build from running?

flake.nix
```diff
@@ -188,6 +188,7 @@
               '';
             }))
             nodePackages.pnpm
+            #nodePackages.node-gyp-build
```
sveitser, line 191: Remove?

Ancient123:
Apparently I hadn't pushed my cleanup changes. Should be good for another review now (also did some of the other build changes)

.github/workflows/arm-build.yaml
```diff
@@ -0,0 +1,123 @@
...
+        name: Cancel Outdated Builds
+        with:
+          access_token: ${{ github.token }}
+
+      - uses: cachix/cachix-action@v10
+        with:
+          name: espresso-systems-private
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - name: Potential broken submodules fix
```
sveitser, line 36: Not important but I think/hope we don't need this anymore.

sveitser (approved):
LGTM

PR #1007, opened 2022-05-05
Include policy in derivation of wrapped asset code.

### LINKED ISSUE #1009: Different wrapped assets should have different asset codes
Currently the derivation of the asset code for wrapped assets only includes the sponsor address and the ERC20 address. This means that if the same Ethereum account tries to sponsor the same ERC20 with different policies, it will create two different asset definitions with the same code. This violates an assumption of the CAP protocol that asset codes are unique, and in particular triggers an internal consistency assertion in the wallet.

jbearer:
This ensures that assets with identical definitions have identical
codes, which upholds invariants expected by the wallet. It allows
the same user to sponsor the same ERC20 twice with different policies,
which previously lead to crashes.

Closes #1009

### DIFF
--- a/contracts/contracts/AssetRegistry.sol
+++ b/contracts/contracts/AssetRegistry.sol
@@ -70,7 +70,7 @@ contract AssetRegistry {
         require(erc20Address != address(0), "Bad asset address");
         require(!isCapeAssetRegistered(newAsset), "Asset already registered");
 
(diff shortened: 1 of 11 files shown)


philippecamacho (approved):
LGTM
Completed the documentation for `_checkForeignAssetCode` and `_computeAssetDescription` in f4522dfdad9dcfc947b8c2a4493ab850dc6e11ef

jbearer:
Thanks @philippecamacho !

PR #1015, opened 2022-05-07
Update seahorse

jbearer:


bfish713:
> Should not we merge [EspressoSystems/seahorse#157](https://github.com/EspressoSystems/seahorse/pull/157) before having CAPE point to the new version of seahorse?

We don't have to wait until https://github.com/EspressoSystems/seahorse/pull/157 is merged for this because Cape is pointing at the release-2.0 branch in seahorse and https://github.com/EspressoSystems/seahorse/pull/156 is already merged.

Ancient123 (approved):

PR #1019, opened 2022-05-10
Set CAPE_RELAYER_EMPTY_BLOCK_INTERVAL_SECS=10 in cape-demo-local

### LINKED ISSUE #1018: Set CAPE_RELAYER_EMPTY_BLOCK_INTERVAL_SECS to a small value for local demo
When demoing CAPE locally it is more convenient to have the relayer clean up the pending deposits queue frequently.
Set the environment variable CAPE_RELAYER_EMPTY_BLOCK_INTERVAL_SECS to 10 seconds CAPE is running on a local blockchain.

philippecamacho:
Closes #1018 .


jbearer:
I wasn't aware of that nonce thing. Unrelated to this issue but isn't it still possible to get unlucky and have the 30s interval fall right in the window when your confirming the transaction? Is this something we should worry about? Could be a bit awkward if it happens during the demo

jbearer (approved):

sveitser:
@philippecamacho @jbearer How about we don't use the same Ethereum wallet for relayer and user? Then this problem should go away. It will require some changes but I think it's worth doing.

philippecamacho:
> I wasn't aware of that nonce thing. **Unrelated to this issue but isn't it still possible to get unlucky and have the 30s interval fall right in the window when your confirming the transaction?** Is this something we should worry about? Could be a bit awkward if it happens during the demo

Indeed this could happen and this is a problem for the demo. Let us go for @sveitser's [solution](https://github.com/EspressoSystems/cape/pull/1019#issuecomment-1123697678).

PR #1022, opened 2022-05-11
Deploy tokens from tokenOwner (2nd account)

sveitser:
Have not tested this with the GUI.

 When using metamask with the test mnemonic, use the 2nd Account (`0x70997970C51812dc3A010C7d01b50e0d17dc79C8`).

Note that the token addresses change to

    CAPE_TOKEN_ADDRESS_DAI=0x948B3c65b89DF0B4894ABE91E6D02FE579834F8F
    CAPE_TOKEN_ADDRESS_SIT=0x8464135c8F25Da09e49BC8782676a84730C318bC    
    CAPE_TOKEN_ADDRESS_WETH=0x71C95911E9a5D330f4D621842EC243EE1343292e

### DIFF
--- a/contracts/deploy/00_cape.ts
+++ b/contracts/deploy/00_cape.ts
@@ -7,11 +7,7 @@
 
 import { HardhatRuntimeEnvironment } from "hardhat/types";
 import { DeployFunction } from "hardhat-deploy/types";
-import { utils, BigNumber } from "ethers";
-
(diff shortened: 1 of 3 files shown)


sveitser:
@philippecamacho we could also make it such that the contract addresses don't change by deploying from account 0 and then transferring the tokens to account 1.

PR #1025, opened 2022-05-11
Use solc from nixpkgs

DieracDelta:
Alias `mySolc` to `nixpkgs#legacyPackages.${system}.solc`. This bumps the solc version to 0.8.13 (was 0.8.10 before) and allows for us to use aarch64. 

Testing: I've tested this on `aarch-linux` with `cargo test --release`.  However, the hivemind I ran this with was for x86_64-linux b/c binfmt looks like it isn't exposing the port to the host VM. @Ancient123 do we have a cluster machine to build and test this on?


Ancient123:
The CI failure is from the git security update we had to fix in other repos as well.
`git config --global --add safe.directory "$PWD"` seems to be the needed magic.

PR #1026, opened 2022-05-11
Use big integers to represent aggregate amounts

### LINKED ISSUE #1016: Use u128 for aggregate amounts in API responses
Due to ERC20s having many decimal places (18 is normal) it is common to have a balance of a wrapped asset type in excess of 2^64. Each individual record is still required by CAP to be at most 2^63, but we currently use `u64` to represent _aggregate_ balances in API responses like `getbalance`. These can lead to overflow and incorrectly reported balances when aggregating a balance across multiple records/accounts.

jbearer:
Closes #1016


pictographer (approved):
LGTM :-)

PR #1030, opened 2022-05-12
Use distinct mnemonics for relayer and faucet

### LINKED ISSUE #1029: Use distinct mnemonics for each service
We're re-using the same mnemonic for the demo and sometimes have transactions fail due to nonce re-use ([zulip thread](https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/Faucet.2Frelayer.20error)).

I think at the least we should use different mnemonics for

- `CAPE_FAUCET_WALLET_MNEMONIC`
- `CAPE_RELAYER_WALLET_MNEMONIC`

and make sure these are **not** used as Metamask mnemonics. Alternatively we could use the same mnemonic but different accounts. The second solution would however require some changes to the rust code.

The script to generate new mnemonics (and show the address of the first account) is `generate-eth-mnemonic`.

```
> generate-eth-mnemonic 
spoon current vehicle student reward behave goose trick acoustic tooth jaguar jewel,0x8572a89eabb0f7D90807DD352D54dc994619FB8a
```

We also need to fund new accounts with Ether. We can add a `fund-wallet` script:

```bash
#!/usr/bin/env bash
set -euxo pipefail

geth --exec "eth.sendTransaction({from: eth.accounts[0], to: \"$1\", value: $2})" \
    attach http://127.0.0.1:8545
```
To fund a wallet with 100 Ether run
```
fund-wallet 0x8572a89eabb0f7D90807DD352D54dc994619FB8a 100e18
```

sveitser:
- Add fund wallet script.

Tested with GUI, the faucet works but I'm having trouble wrapping an
asset. The page hangs after submitting the wrapping tx through metamask.

Edit: It looks like my metamask was in a weird state and not properly connected. After re-connecting the wrapping works.
Close #1029

### DIFF
--- a/bin/cape-demo-local
+++ b/bin/cape-demo-local
@@ -51,11 +51,11 @@ export CAPE_EQS_NUM_CONFIRMATIONS=1 # More responsive local demo.
 
 export CAPE_RELAYER_PORT=50020
 export CAPE_RELAYER_URL="http://localhost:$CAPE_RELAYER_PORT"
-export CAPE_RELAYER_WALLET_MNEMONIC="$TEST_MNEMONIC"
+export CAPE_RELAYER_WALLET_MNEMONIC="organ task tattoo favorite salon effort matrix main original guitar drastic elder"
 
 export CAPE_FAUCET_PORT=50030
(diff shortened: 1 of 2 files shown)


philippecamacho (approved):
LGTM

PR #1031, opened 2022-05-12
Include pending txns nonce count for local demo

sveitser:
This will allow the relayer to submit more than one tx per block.


philippecamacho (approved):
LGTM
I tested this change against the [latest GUI](https://github.com/EspressoSystems/cape-ui/commit/5070b1b4aef260f4155f847acbc3e0e8343ebeb9) andi did not have any problem.

PR #1035, opened 2022-05-12
Relayer sends empty block only if the pending deposits queue is not e…

### LINKED ISSUE #1033: Relayer sends an empty block only if the pending deposits queue is non empty
See discussion https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/Faucet.2Frelayer.20error/near/282090994

philippecamacho:
…mpty.

Closes #1033


sveitser (approved):
LGTM but I think we should add a bit of logging so that we can tell what are relayed transactions and what are empty blocks based on the logs.

PR #1036, opened 2022-05-12
Allow cross-origin JSON-RPC requests for local demo

jbearer:
I ran through the full demo script with this change, no issues, so at least it doesn't make things worse. Not sure how to test that it actually fixes the CORS issue but according to [the docs](https://geth.ethereum.org/docs/rpc/server#http-server) it should.


philippecamacho (approved):
LGTM

PR #1037, opened 2022-05-12
Set number of decimals for all local demo tokens to 6.

philippecamacho:


jbearer (approved):

sveitser:
Was able to successfully wrap 123 tokens in the GUI but the displayed balance is 1.23e-10 so the computation somewhere still assumes 18 decimals.

image

PR #1040, opened 2022-05-12
Revert "Use big integers to represent aggregate amounts (#1026)"

jbearer:
This reverts commit 2ceec5e9b9931787d8d68a6d1dbd0bdacc290c0a.

This commit changed the serialization of balances from numbers to
strings, since they can now exceed the capacity of a JSON number.
This caused a problem in the GUI where it was aggregating balances,
where it ended up concatenating strings rather than adding integers.

Since it is a nontrivial change to fix this in the GUI, with the demo
tomorrow, I am reverting this commit for now. Next week, we should
either fix the GUI to handle strings properly, or just do the balance
aggregation in the backend, or both. After that we can commit this
again.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5202,8 +5202,8 @@ dependencies = [
 
 [[package]]
 name = "seahorse"
-version = "0.2.3"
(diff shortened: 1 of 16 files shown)


pictographer (approved):
Sad to see the old code go, but understandable.

PR #1044, opened 2022-05-16
Include pending transaction in nonce count by default

### LINKED ISSUE #1032: Include pending txns in nonce computation (for goerli enviornment)
In #1031 we changed this for the local demo backend to reduce the chances of race conditions. We should eventually also change it for the goerli deployment. Otherwise, according to my understanding, the relayer will only be able to submit a single transaction per block.

sveitser:
This allows the relayer to include more than one Ethereum txn per block.

However it does not fix all race condition and submitting a new
transaction before the previous one is in the mempool will still lead to
an error.

We could use ethers-rs' nonce manager middleware to perform the nonce
count locally but this also present a some new challenges like handling
the case where the count is out of sync.

Close #1032


sveitser:
@jbearer Yes. If we want to avoid errors due to these race conditions popping up in the front-end we do have to do some extra work. I'll add it to the discussion list.

PR #1046, opened 2022-05-16
Include aggregate balances in getbalance/all.

### LINKED ISSUE #1039: Aggregate balances in backend
`getbalance` should return an aggregated balance for each asset type, in addition to the balance for each account. Currently the GUI does the aggregation itself, which is both slow and a source of bugs (for example, changing the serialization format for balances from `number` to `string` representing an `BigInteger` broke the UI, since `+` in JS means both "add" and "concatenate".

jbearer:
This also restructures the output of getbalance/all to decouple the
asset metadata from the balances themselves, to avoid duplication of
metadata for each asset type.

Closes #1039

### DIFF
--- a/wallet/src/ui.rs
+++ b/wallet/src/ui.rs
@@ -371,7 +371,7 @@ pub enum PrivateKey {
 pub enum Balances {
     /// The balance of a single asset, in a single account.
     One(u64),
-    /// All the balances of an account, by asset type with asset info.
+    /// All the balances of an account, by asset type.
     Account(HashMap<AssetCode, u64>),
     /// All the balances of all accounts owned by the wallet.
     All {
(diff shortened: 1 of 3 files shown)


wallet/src/ui.rs
```diff
@@ -359,13 +368,40 @@ pub enum PrivateKey {
 }
 
 #[derive(Clone, Debug, PartialEq, Serialize, Deserialize)]
-pub enum BalanceInfo {
+pub enum Balances {
     /// The balance of a single asset, in a single account.
-    Balance(u64),
+    One(u64),
     /// All the balances of an account, by asset type with asset info.
```
shenkeyao, line 374: Nit: Remove "with asset info".

wallet/src/ui.rs
```diff
@@ -359,13 +368,40 @@ pub enum PrivateKey {
...
+            Self::One(_) => Box::new(empty()),
+            Self::Account(by_asset) => Box::new(by_asset.keys()),
+            Self::All {
+                by_account,
+                aggregate,
+            } => Box::new(
+                by_account
+                    .values()
+                    .flat_map(|by_asset| by_asset.keys())
+                    .chain(aggregate.keys()),
```
shenkeyao, line 395: Isn't each `AssetCode` in `aggregate` guaranteed to be also included in `by_account`? I'm wondering why we need to add both `aggregate.keys()` and `by_asset.keys()`. Can we just use `aggregate` to get the assets?

shenkeyao (changes requested):
Looks good. Just a comment about the `assets()` function.

wallet/src/ui.rs
jbearer: Good point. I used `by_account` since that's the primary source of data, it felt more appropriate to derive everything from that. Got rid of the redundant `.chain(...)`

shenkeyao (approved):

PR #1058, opened 2022-05-18
Problem: relayer Eth txn easily runs out of gas

### LINKED ISSUE #1056: Set manual gas limit in relayer to avoid out of gas errors
6M should be ok.

sveitser:
Solution: specify the gas limit manually.

Adds a `CAPE_RELAYER_GAS_LIMIT` env var so we can adjust the value
without changing code if needed.

It would be better to create a custom type for the gas limit to avoid
confusing it with other u64s. It may also we worth it to group all the
Ethereum config together instead of passing around so many variables.
In light of time pressure I went with what was quicker to implement.

Close #1056

### DIFF
--- a/contracts/rust/src/cape/events.rs
+++ b/contracts/rust/src/cape/events.rs
@@ -80,6 +80,7 @@ mod tests {
             &connection.contract,
             block_with_memos.clone(),
             BlockNumber::Latest,
+            10_000_000, // gas limit
         )
         .await?
         .await?;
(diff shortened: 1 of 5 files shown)


contracts/rust/src/cape/events.rs
```diff
@@ -80,6 +80,7 @@ mod tests {
             &connection.contract,
             block_with_memos.clone(),
             BlockNumber::Latest,
+            10_000_000, // gas limit
```
philippecamacho, line 83: Why do we have a magic number here? Should not it be `DEFAULT_RELAYER_GAS_LIMIT` instead?

philippecamacho (approved):
LGTM. Left a question.

contracts/rust/src/cape/events.rs
sveitser: Yes but that one is defined in the relayer crate.

PR #1060, opened 2022-05-18
Feat/t1059 relayer sequential block submission

### LINKED ISSUE #1059: Relayer: submit blocks in sequence
Submitting the blocks one by one should avoid most race conditions affecting the relayer block submissions.

sveitser:
Note: Only a97d268009f2489a181bf7ba7dc39fd9c37fed98 is part of this PR.

I haven't used the rust Mutex before so would especially appreciate if someone could take a look at that part of the code. I've been trying to test with the local cape demo but am currently having issues where the GUI can't fetch balances. 

Edit: The GUI errors are due to the changes in https://github.com/EspressoSystems/cape/commit/4596705abaf26dfc58e411722f032a2595e29abe.

Close #1059

### DIFF
--- a/relayer/src/lib.rs
+++ b/relayer/src/lib.rs
@@ -70,15 +70,15 @@ fn server_error<E: Into<Error>>(err: E) -> tide::Error {
 }
 
 #[derive(Clone)]
-struct WebState {
+pub struct WebState {
     contract: CAPE<EthMiddleware>,
     nonce_count_rule: NonceCountRule,
     gas_limit: u64,
-    block_submission_mutex: Arc<Mutex<u64>>,
+    block_submission_mutex: Arc<Mutex<()>>,
 }
 
 impl WebState {
-    fn new(
+    pub fn new(
         contract: CAPE<EthMiddleware>,
         nonce_count_rule: NonceCountRule,
         gas_limit: u64,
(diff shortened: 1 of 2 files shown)


relayer/src/lib.rs
```diff
@@ -231,20 +237,22 @@ pub async fn submit_empty_block_loop(
     gas_limit: u64,
     empty_block_interval: Duration,
 ) -> Result<(), Error> {
+    let web_state = WebState::new(contract, nonce_count_rule, gas_limit);
```
jbearer, line 240: This will create a new `Mutex` which doesn't synchronize with the one used in the request handler. I think this function should take a (_the_) `WebState` as an argument.

relayer/src/lib.rs
```diff
@@ -73,6 +74,22 @@ struct WebState {
     contract: CAPE<EthMiddleware>,
     nonce_count_rule: NonceCountRule,
     gas_limit: u64,
+    block_submission_mutex: Arc<Mutex<u64>>,
```
jbearer, line 77: `Arc<Mutex<()>>` is preferred if we're not actually using the value in the mutex.

jbearer (changes requested):

jbearer (approved):

PR #1063, opened 2022-05-18
Run multiple random wallet instances against local backend

sveitser:
This (commit 5a3da09427e27c5206dd08d77dcba49a99e5c17d) is an experiment to see if we can run multiple random wallet test instances against the local demo.

It almost works a bit too smoothly. I think in order to make it useful we need to simulate more realistic network conditions with higher latency and packet drops.

### DIFF
--- a/bin/run-random-wallet
+++ b/bin/run-random-wallet
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+# Copyright (c) 2022 Espresso Systems (espressosys.com)
+# This file is part of the Configurable Asset Privacy for Ethereum (CAPE) library.
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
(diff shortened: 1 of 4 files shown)


philippecamacho:
Looks awesome! So the only obstacle to run it against the CAPE system deployed on Goerli is the shortage of Goerli eth?

sveitser:
@philippecamacho Sorry just saw your comment. The tokens are deployed inside the random-wallet test so that should work. However the ether funding uses the unlocked account on the geth node (and funds new wallets with 1000 Eth) so will require some adjustments.
 
What already works and I think could be quite useful is to run it locally and introduce suboptimal network conditions. For example to delay requests by 10 +/- 1000 msec with 25% correlation (I think).

```
sudo tc qdisc add dev lo root handle 1: netem delay 10msec 1000msec 25%
```

One can run `ping localhost` to get a an idea of what this does to the network and then just run the random wallet tests to see if the relayer can handle it.

To restore the normal conditions
```
sudo tc qdisc del dev lo root
```

Issue #1062, opened 2022-05-18
Consider using a "valid_until_field" for dealing with pending transactions in the Wallet


philippecamacho:
See https://www.notion.so/espressosys/sync-e8e4cf71c2214015a79bd4499b198118#178190b5a9294aedb4c48aba0b8c7772

PR #1066, opened 2022-05-18
Set the number of roots in the deployed CAPE contract to 40.

philippecamacho:


jbearer (approved):
So are we not going to add a `valid_until` thing? Or is this just a stop-gap?

philippecamacho:
> So are we not going to add a `valid_until` thing? Or is this just a stop-gap?

In the short term we will just stick to the simplest solution. Another [ticket](https://github.com/EspressoSystems/cape/issues/1062) has been created for the "valid_until" approach.

PR #1067, opened 2022-05-19
EQS changes to make jsonrpc connection optional

kmchicoine:
Added HashMap<AssetDefinition, Address> and endpoint for wallet


eqs/src/routes.rs
```diff
@@ -178,6 +180,22 @@ pub async fn get_transaction_by_hash(
...
+        .address_from_assetdef
+        .get(&bindings[":assetdef"].value.to::<AssetDefinition>()?)
+        .cloned()
+    {
+        Ok(Some(addr))
+    } else {
+        Ok(None)
+    }
+}
+
```
kmchicoine, line 198: Realizing now that this doesn't check for the wallet asking for a bad/nonexistent AssetDefinition. Could be solved by adding <native AssetDefinition, None> to the address_from_assetdef map- could this be done during the eth_polling new function?

eqs/api/api.toml
```diff
@@ -68,3 +68,8 @@ ERROR_hash = "A valid commitment hash is required. Commitment begin with CMTMNT_
 [route.healthcheck]
 PATH = [ "healthcheck" ]
 DOC = "Responds with JSON {\"status\": \"available\"}."
+
+[route.get_contract_address]
+PATH = [ "get_address/:assetdef" ]
```
jbearer, line 73: IIRC the route name has to be the same as the first segment of the path. Let's make both of these names more descriptive, like `get_wrapped_erc20_address`

eqs/api/api.toml
```diff
@@ -68,3 +68,8 @@ ERROR_hash = "A valid commitment hash is required. Commitment begin with CMTMNT_
...
+":assetdef" = "AssetDefinition"
```
jbearer, line 74: I don't think `AssetDefinition` is a type recognized by our request parsing stuff (yet). I think this would have to be `TaggedBase64` and then you'll deserialize it.

eqs/src/eth_polling.rs
```diff
@@ -365,6 +370,19 @@ impl EthPolling {
...
+                    //add assets to local map
+                    for wrap in wraps {
+                        match wrap {
+                            CapeTransition::Wrap { ro, src_addr, .. } => {
+                                updated_state
+                                    .address_from_assetdef
+                                    .insert(ro.asset_def, Address::from(src_addr));
+                            }
+                            CapeTransition::Faucet { .. } => {}
+                            CapeTransition::Transaction { .. } => {}
```
jbearer, line 382: If these are unreachable (assuming based on the name `wraps`) can we replace this with `_ => unreachable!()` or `panic!()` or something?

eqs/src/eth_polling.rs
```diff
@@ -365,6 +370,19 @@ impl EthPolling {
                     updated_state.last_reported_index = Some(current_index);
                     self.last_event_index = Some(current_index);
 
+                    //add assets to local map
+                    for wrap in wraps {
+                        match wrap {
+                            CapeTransition::Wrap { ro, src_addr, .. } => {
+                                updated_state
+                                    .address_from_assetdef
+                                    .insert(ro.asset_def, Address::from(src_addr));
```
jbearer, line 379: I believe `src_addr` here is the wrapper's Ethereum address and not the ERC20 contract address. I think you want the `erc20_code` field.

eqs/src/routes.rs
jbearer: Not sure what you mean. If the wallet asks for an `AssetDefinition` that isn't in the mapping, this will return `None`, which is the correct behavior. FWIW though I think you could eliminate the `if` statement and just return `Ok(query_result_state....cloned())`

eqs/src/eth_polling.rs
```diff
@@ -365,6 +370,19 @@ impl EthPolling {
                     updated_state.last_reported_index = Some(current_index);
                     self.last_event_index = Some(current_index);
 
+                    //add assets to local map
+                    for wrap in wraps {
```
jbearer, line 374: It would be better to update this mapping on sponsor, rather than wrap, since sponsor happens exactly once per asset and it's the operation that actually establishes the mapping. Currently, though, there is no event for sponsors. Since we're going to be redeploying the contract soon anyway (to handle 128 bit amounts) should we also add an event for sponsors? cc @philippecamacho  @sveitser

jbearer (changes requested):

eqs/src/eth_polling.rs
philippecamacho: I agree, I created a [ticket](https://github.com/EspressoSystems/cape/issues/1068) so that we can come back to it later.

jbearer (approved):

PR #1070, opened 2022-05-19
Improvements to the faucet service

jbearer:
* A background thread breaks large records into smaller ones,
  ensuring that the faucet wallet always has at least a given
  number of records at a time. This should help in processing
  simultaneous requests.
* Faucet grants are distributed in multiple records, so that the
  wallet receiving the grant can easily use the records it receives
  to carry out simultaneous transactions.
* The faucet waits until it has finished scanning the ledger before
  it starts serving requests.


faucet/src/faucet.rs
```diff
@@ -170,27 +219,207 @@ async fn request_fee_assets(
...
+
+/// Break large records into smaller records.
+///
+/// When signalled on `wakeup`, this thread will break large records into small records of size
+/// `state.grant_size`, until there are at least `state.num_records` distinct records in the wallet.
+async fn break_up_records(state: FaucetState, mut wakeup: mpsc::Receiver<()>) {
+    loop {
+        // Wait until we have few enough records that we need to break them up.
+        //
+        // This is a simulation of a condvar loop, since async condvar is unstable, hence the manual
```
nyospe, line 301: Thank you for explaining this here, I spent a couple of minutes scratching my head until I noticed the comment

nyospe (approved):
LGTM

jbearer:
When I test this with the GUI, I get an error toast after a faucet request, even though the actual POST succeeds and the assets show up in my wallet. This is because of the way the GUI interprets the faucet's return type:
```typescript
const res = await faucet(newAddress);
      if (res.status === 200) {
        const data = await res.json();
        if (!data) {
          toast.success('Success');
          setCapeReceived(true);
        } else {
          toast.error('An error occurred. Please try again later.');
          return;
        }
      } else {
        toast.error('An error occurred. Please try again later.');
      }
```

`if (!data)` means the GUI interprets _any_ nonempty response payload as an error message.

I want to keep the new API where the faucet returns the transaction receipts. This is nice for debuggability, plus we may be able to use it later on to display the pending faucet transactions in the GUI, which would greatly improve UX. I think we should keep this PR as is and fix this on the UI side, but I will wait until after tomorrow's bug bash to merge it.

Issue #1068, opened 2022-05-19
Create a sponsor event when `AssetRegistry.sponsorCapeAsset` is called


philippecamacho:
See discussion https://github.com/EspressoSystems/cape/pull/1067#discussion_r877322595

philippecamacho:
Close in https://github.com/EspressoSystems/cape/pull/1072.

PR #1071, opened 2022-05-20
Make the web3 connection in the wallet backend optional.

jbearer:
The wallet now uses the EQS to get the wrapped ERC20 code for an
asset. Other functions that require a contract connection, like
sponsor and wrap, are optional and will not work if a web3 endpoint
is not provided. These functions are not required for the GUI.


nyospe (approved):
Straightforward. LGTM.

jbearer:
Will not work until we fix #1068 but should be ready to go once that's in (Philippe is working on it)

PR #1072, opened 2022-05-20
Emit an event in the CAPE contract when an asset is sponsored.

philippecamacho:


jbearer (approved):
This oughta work. We'll have to change the mapping in the EQS from `<AssetDefinition, Address>` to `<AssetCode, Address>` and change the wallet to query that way. But that's probably a better interface anyways since keying on `AssetCode` is strictly more permissive

PR #1087, opened 2022-05-25
Wait for sponsored asset to be reflected in EQS

shenkeyao:
- Adds backend functions to wait until the sponsored asset is reflected in the EQS.
- Checks the EQS is running in `buildsponsor`.
- Adds the `waitforsponsor` endpoint and updates the test.

Related issue: https://github.com/EspressoSystems/cape-ui/issues/349.


wallet/src/backend.rs
```diff
@@ -538,6 +534,33 @@ impl<'a, Meta: Serialize + DeserializeOwned + Send> CapeWalletBackend<'a>
         Ok(address.map(Erc20Code::from))
     }
 
+    async fn wait_for_wrapped_erc20_code(
+        &mut self,
+        asset: &AssetDefinition,
```
jbearer, line 539: We should change this to take an `AssetCode`, makes it easier to call

wallet/src/wallet-api/main.rs
```diff
@@ -1021,6 +1023,13 @@ mod tests {
         let view_amount = true;
         let view_address = false;
 
+        // Start the EQS.
+        let mut rng = ChaChaRng::from_seed([1u8; 32]);
+        let universal_param = &*UNIVERSAL_PARAM;
+        let (_, _, _, contract_address, _) =
+            create_test_network(&mut rng, universal_param, None).await;
+        let _ = spawn_eqs(contract_address, Some(50087)).await;
+
```
jbearer, line 1032: We decided that starting the real EQS here doesn't do anything, right?

wallet/src/wallet-api/routes.rs
```diff
@@ -744,9 +744,16 @@ async fn newasset(
...
 async fn buildsponsor(
+    options: &NodeOpt,
     bindings: &HashMap<String, RouteBinding>,
     wallet: &mut Option<Wallet>,
 ) -> Result<(sol::AssetDefinition, String), tide::Error> {
+    // Make sure the EQS is running.
+    surf::connect(options.eqs_url())
+        .send()
+        .await
+        .expect("Cannot build the sponsor transaction since EQS is down");
```
jbearer, line 755: Ah, this is why the tests need the EQS to be running. Seems like this is another thing that should be abstracted by `CapeWalletBackend`. It doesn't make much sense for the tests to ping the real EQS when the mock backend is talking to a mock query service. Not a big deal, though, since it only amounts to a slight inefficiency in test code.

wallet/src/wallet.rs
```diff
@@ -45,6 +45,14 @@ pub trait CapeWalletBackend<'a>: WalletBackend<'a, CapeLedger> {
         asset: &AssetDefinition,
     ) -> Result<Option<Erc20Code>, CapeWalletError>;
 
+    /// Wait until the EQS reflects the associated ERC20 code of the CAPE asset, or the `timeout`
+    /// (in seconds) is reached.
+    async fn wait_for_wrapped_erc20_code(
+        &mut self,
+        asset: &AssetDefinition,
+        timeout: Option<u64>,
```
jbearer, line 53: We should probably be taking a `Duration` here and converting seconds to `Duration` at a higher level (like the API route handler)

wallet/src/wallet-api/routes.rs
shenkeyao: Yeah, it seems better to ping the EQS by `CapeWalletBackend` rather than `buildsponsor`. I'll modify this.

wallet/src/backend.rs
shenkeyao: For the mock backend, `wait_for_wrapped_erc20_code` calls `get_wrapped_erc20_code` which takes an `AssetDefinition` to check if it's in the `erc20_registrar` table. If we're going to modify `wait_for_wrapped_erc20_code` to take an `AssetCode`, then we also need to change `erc20_registrar` and some other places (e.g., the `wrapped_erc20_registrar` field of `CapeContract`). I guess those structs/data take an `AssetDefinition` rather than `AssetCode` for some reason, so I'm not sure if we should change the type.

jbearer: We can add the `AssetCode` -> `Erc20Code` mapping to the mock EQS in `mocks.rs` and add the `AssetSponsored` event in `model.rs` and basically simulate exactly what the real EQS does. But TBH it's probably fine to do a linear search for an `AssetDefinition` with the given code for now, in the mock backend. Something like `erc20_registrar.iter().find(|definition| definition.code == code)`

jbearer (approved):

PR #1088, opened 2022-05-26
Feat/t1017 circuit 128 bit followup (v2)

sveitser:
- Remove some unnecessary type conversions.
- Use u64 for small uints instead of u128.
- Replace magic value `1000` with `DEFAULT_WRAPPED_AMOUNT`

### DIFF
--- a/wallet/src/wallet-api/web.rs
+++ b/wallet/src/wallet-api/web.rs
@@ -35,7 +35,6 @@ use tide::{
 };
 
 pub const DEFAULT_ETH_ADDR: Address = H160([2; 20]);
-// TODO change these to Amount type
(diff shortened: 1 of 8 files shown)


wallet/src/wallet-api/web.rs
```diff
@@ -35,7 +35,6 @@ use tide::{
 };
 
 pub const DEFAULT_ETH_ADDR: Address = H160([2; 20]);
-// TODO change these to Amount type
```
sveitser, line 38: Compiler says "no": `RecordAmount::from<u128>` is not a constant function.

jbearer (approved):

PR #1104, opened 2022-06-01
Remove versioned name of nix cache docker volume

Ancient123:


sveitser:
Looked like the previous build was failing because the an executable wasn't built when we tried to run in. The order of these steps has been swapped on `main`.

PR #1107, opened 2022-06-03
Fix contract addresses for local demo

sveitser:
The previous ones were wrong, probably because I didn't have an empty
chain when I deployed.


philippecamacho (approved):
LGTM

Issue #1115, opened 2022-06-07
Serialize amounts as strings


philippecamacho:

philippecamacho:
See https://www.notion.so/espressosys/sync-4c9de28e08c7438b9a01bbf4c7474a2c#e3a94bfde7664c6bb1fe8638435c6892

PR #1118, opened 2022-06-07
Change RecordAmount to String for UI

shenkeyao:
Serializes amount fields in `ui.rs` as `String` rather than `RecordAmount`.

Closes https://github.com/EspressoSystems/cape/issues/1115.

### DIFF
--- a/wallet/src/ui.rs
+++ b/wallet/src/ui.rs
@@ -18,14 +18,13 @@ use jf_cap::{
     structs::{AssetCode, AssetDefinition as JfAssetDefinition, AssetPolicy as JfAssetPolicy},
 };
 use net::UserAddress;
-use num_traits::identities::Zero;
 use reef::cap;
 use seahorse::{
     accounts::{AccountInfo, KeyPair},
     asset_library::Icon,
     events::EventIndex,
     txn_builder::RecordInfo,
-    MintInfo, RecordAmount,
+    MintInfo,
 };
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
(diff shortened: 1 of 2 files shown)


jbearer:
What about the other uses of `RecordAmount` in API response? E.g.
* `AssetDefinition::viewing_threshold`
* `Record`
* `sol::AssetPolicy`
* `sol::RecordOpening`

Feel free to just changes those to `String` as well in the interests of time. But long term I think it might be cleaner to change the `serde` impls for `RecordAmount` in Seahorse, using e.g.
```rust
#[serde_as]
pub struct RecordAmount(#[serde_as(as = "DisplayFromStr")] pub Amount);
```
(see https://docs.rs/serde_with/1.14.0/serde_with/rust/display_fromstr/index.html)

shenkeyao:
> What about the other uses of `RecordAmount` in API response? E.g.
> 
> * `AssetDefinition::viewing_threshold`
> * `Record`
> * `sol::AssetPolicy`
> * `sol::RecordOpening`
> 
> Feel free to just changes those to `String` as well in the interests of time. But long term I think it might be cleaner to change the `serde` impls for `RecordAmount` in Seahorse, using e.g.
> 
> ```rust
> #[serde_as]
> pub struct RecordAmount(#[serde_as(as = "DisplayFromStr")] pub Amount);
> ```
> 
> (see https://docs.rs/serde_with/1.14.0/serde_with/rust/display_fromstr/index.html)

Oops I missed those places. Will change them as well.

I'm going to use `String` for now to resolve the UI issue asap, but will file an issue to change the impl in Seahorse.

jbearer (approved):

shenkeyao:
@jbearer While modifying those amount fields, though fortunately for this PR there are not many `RecordAmount` involved, I think it's definitely better to change the `serde` implementation in Seahorse than use `String` in Cape. I added an [issue](https://github.com/EspressoSystems/seahorse/issues/187) for this to be addressed later.

PR #1130, opened 2022-06-09
In local demo, wait for faucet to become available before using it

jbearer:


sveitser (approved):
LGTM and worked well for me.

PR #1139, opened 2022-06-13
Write a dummy entry to storage during healthcheck

sveitser:


sveitser:
The new test fails but works locally. :unamused:

sveitser:
2 out of 2 failures on CI. I wonder if it's a timing issue or if the removing the write permissions on the directory just doesn't work on the CI.

address_book/src/lib.rs
```diff
@@ -65,11 +65,16 @@ impl FileStore {
 impl Store for FileStore {
     fn save(&self, address: &UserAddress, pub_key: &UserPubKey) -> Result<(), std::io::Error> {
         let tmp_path = self.tmp_path(address);
-        fs::write(
+        match fs::write(
```
nyospe, line 68: Good!

address_book/src/lib.rs
```diff
@@ -247,9 +252,14 @@ async fn request_pubkey<T: Store>(
 /// normally, a response with status 503 and payload {"status":
 /// "unavailable"} should be added.
 async fn healthcheck<T: Store>(
-    mut _req: tide::Request<ServerState<T>>,
+    req: tide::Request<ServerState<T>>,
 ) -> Result<tide::Response, tide::Error> {
-    Ok(tide::Response::builder(200)
+    // Ensure we can write to storage.
```
nyospe, line 257: Certainly *looks* like the right thing to do here...

nyospe (approved):
Not sure what's up in CI, but this looks like it should be correct...

PR #1157, opened 2022-06-21
Asynchronous faucet.

### LINKED ISSUE #1151: Improve scalability of Faucet
The faucet cannot currently handle too many concurrent requests.

See  [Faucet overloaded](https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/Faucet.20overloaded/near/286380634).

jbearer:
Instead of processing requests directly in `tide` request handlers,
    the request handler simply adds the request to a queue, and the
    requests are dequeued and processed by a fixed number of worker
    threads.
    
    This solves two problems:
    * Requests taking a long time to process do not time out the HTTP
      request
    * We limit the CPU usage by limiting the number of worker threads,
      so the server does not get overloaded.
    
    Closes #1151

### DIFF
--- a/faucet/src/faucet.rs
+++ b/faucet/src/faucet.rs
@@ -10,6 +10,7 @@
 
 extern crate cape_wallet;
 
+use async_channel as mpmc;
 use async_std::{
     sync::{Arc, Mutex, RwLock},
     task::{sleep, spawn, JoinHandle},
@@ -21,7 +22,6 @@ use cape_wallet::{
     wallet::{CapeWallet, CapeWalletError},
 };
 use futures::{channel::mpsc, SinkExt, StreamExt};
-use itertools::Itertools;
 use jf_cap::{
     keys::{UserKeyPair, UserPubKey},
     structs::{AssetCode, FreezeFlag},
@@ -36,6 +36,7 @@ use seahorse::{
 };
 use serde::{Deserialize, Serialize};
 use snafu::Snafu;
+use std::collections::HashSet;
 use std::path::PathBuf;
 use std::time::Duration;
 use structopt::StructOpt;
@@ -122,6 +123,19 @@ pub struct FaucetOptions {
     /// Minimum amount of time to wait between polling requests to EQS.
     #[structopt(long, env = "CAPE_WALLET_MIN_POLLING_DELAY", default_value = "500")]
     pub min_polling_delay_ms: u64,
+
+    /// Maximum number of outstanding requests to allow in the queue.
+    ///
+    /// If not provided, the queue can grow arbitrarily large.
+    #[structopt(long, env = "CAPE_FAUCET_MAX_QUEUE_LENGTH")]
+    pub max_queue_len: Option<usize>,
+
+    /// Number of worker threads.
+    ///
+    /// It is a good idea to configure the faucet so that this is the same as
+    /// `num_records / num_grants`.
+    #[structopt(long, env = "CAPE_FAUCET_NUM_WORKERS", default_value = "5")]
+    pub num_workers: usize,
 }
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq)]
@@ -135,6 +149,7 @@ pub enum FaucetStatus {
 struct FaucetState {
     wallet: Arc<Mutex<CapeWallet<'static, CapeBackend<'static>>>>,
     status: Arc<RwLock<FaucetStatus>>,
+    queue: FaucetQueue,
     grant_size: RecordAmount,
     num_grants: usize,
     fee_size: RecordAmount,
@@ -157,6 +172,7 @@ impl FaucetState {
         Self {
             wallet: Arc::new(Mutex::new(wallet)),
             status: Arc::new(RwLock::new(FaucetStatus::Initializing)),
+            queue: FaucetQueue::new(opt.max_queue_len),
             grant_size: opt.grant_size,
             num_grants: opt.num_grants,
             fee_size: opt.fee_size,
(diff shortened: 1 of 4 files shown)


jbearer:
Note that we aren't currently getting parallelism out of having multiple workers, since each worker currently locks the global wallet exclusively while it is transferring. So for now it doesn't make much sense to configure `num_workers` greater than 1. But this design makes it pretty easy to get even fancier, like having `n` separate wallets for `n` workers, achieving true parallelism up to the constraints of the hardware.

faucet/src/faucet.rs
```diff
@@ -202,6 +244,225 @@ pub fn faucet_error(source: CapeWalletError) -> tide::Error {
...
+        // channel, or if we have encountered a `None` entry for it and skipped it.
+        let mut processed = HashSet::new();
+        // We are encountering requests in reverse order, so if we need to add them to the queue, we
+        // will add them to this [Vec] and then reverse it at the end before adding them to the
+        // message channel.
+        let mut queue = Vec::new();
+        let entries: Vec<(UserPubKey, Option<usize>)> =
+            persistent_queue.iter().collect::<Result<_, _>>()?;
+        for (key, val) in entries.into_iter().rev() {
+            if !index.contains_key(&key) {
```
sveitser, line 372: I think this is the same as `index.insert(key.clone(), val);`, so we could also write it in the more compact version.

faucet/src/faucet.rs
```diff
@@ -202,6 +244,225 @@ pub fn faucet_error(source: CapeWalletError) -> tide::Error {
...
+                if let Some(val) = val {
+                    // This is the most recent value for `key`, and it is an insert, which means
+                    // `key` is in the queue. Go ahead and add it to the index and the message
+                    // channel.
+                    index.insert(key.clone(), Some(val));
+                } else {
+                    // This is the most recent value for `key`, and it is a delete, which means
+                    // `key` is not in the queue. Remember this information in `index`.
+                    index.insert(key.clone(), None);
+                }
```
sveitser, line 382: I think this is the same as `index.insert(key.clone(), val);`, so we could also write it in the more compact version.

faucet/src/faucet.rs
```diff
@@ -202,6 +244,225 @@ pub fn faucet_error(source: CapeWalletError) -> tide::Error {
...
+/// completes a transfer to the public key, it will call [FaucetQueue::grant], which increments the
+/// counter associated with that public key, persists the change, and instructs the worker to
+/// either continue transferring to the same key or to move on to the next key.
+///
+/// The queue is persistent, so that if the faucet crashes or gets restarted, it doesn't lose the
+/// queue of pending requests. The persistent queue is represented as a log of index entries, of the
+/// form `UserPubKey -> Option<usize>`. An entry `key -> Some(n)` corresponds to updating the
+/// counter associated with `key` to `n`. An entry `key -> None` corresponds to deleting the entry
+/// for `key`. We can recover the in-memory index by simply replaying each log entry and inserting
+/// or deleting into a `HashMap` as indicated.
```
sveitser, line 267: It wasn't too hard to understand the code thanks to the explanations here but I wonder if it would make the code more clear by itself if instead of using `Some(0), Some(n > 0), None` to represent the state of a faucet request we had an enum for that.

sveitser:
Weird. I get 2 faucet test binaries so the new test runs twice. Also seems to happen on the CI.

philippecamacho:
I got some error when running the tests, but it might not be related to the change of this PR.

```
test backend::test::test_anonymous_erc20_transfer ... FAILED <69.927s>

failures:

---- backend::test::test_anonymous_erc20_transfer stdout ----
Sent funding tx to deployer
xfr: 987 bytes
mint: 971 bytes
freeze: 979 bytes
Sent funding tx to deployer
thread 'backend::test::test_anonymous_erc20_transfer' panicked at 'called `Result::unwrap()` on an `Err` value: Failed { msg: "relayer error: response body fails to deserialize: io error: unexpected end of file" }', wallet/src/backend.rs:995:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/result.rs:1749:5
   3: cape_wallet::backend::test::test_anonymous_erc20_transfer::{{closure}}
   4: std::thread::local::LocalKey<T>::with
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   6: async_io::driver::block_on
   7: std::thread::local::LocalKey<T>::with
   8: std::thread::local::LocalKey<T>::with
   9: async_std::task::builder::Builder::blocking
  10: core::ops::function::FnOnce::call_once
  11: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    backend::test::test_anonymous_erc20_transfer

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 69.93s

error: test failed, to rerun pass '-p cape_wallet --lib'
Sending HUP signal to run-geth: 14774
Cleaning up geth node
Removing geth data dir /run/user/1000/cap-ethereum-data-M56icRDG

```

faucet/src/faucet.rs
jbearer: Oh, good point. There used to be some extra logic in one of the branches, but I removed it

faucet/src/faucet.rs
jbearer: Yep, good suggestion

jbearer:
@philippecamacho that _appears_ unrelated but I've also never seen that error before. Can you checkout `main` and see if you get a similar error? I'll try running this test locally a few times as well before I merge this

sveitser:
For me all tests pass.

sveitser:
One thing I noticed is that sometimes the faucet command from the test does not seem to terminate after the tests are over.

jbearer:
I'm getting the same error on `main`. Not sure what's going on but I will go ahead and merge this and then try to figure it out

PR #1158, opened 2022-06-22
Decrease maximum wallet polling delay.

jbearer:
This delay is meant to decrease pressure on the EQS. We have
observed that even at some scale, the EQS uses barely any of its
allotted CPU. Therefore, we can poll more frequently, which will
lead to the "CAPE Syncing" spinner in the UI appearing for shorter
durations, and ultimately making the UI feel just a bit more
responsive.


philippecamacho (approved):
LGTM

PR #1163, opened 2022-06-22
Add wallet-cli to wallet docker image

### LINKED ISSUE #1149: Include `wallet-cli` binary in wallet Docker, or create `wallet-cli` Docker
I need some way for users to get the `wallet-cli` binary, without making them build it themselves, so I can provide instructions for trying out the viewing and freezing capabilities.

sveitser:
Close #1149

### DIFF
--- a/wallet.Dockerfile
+++ b/wallet.Dockerfile
@@ -9,6 +9,7 @@ COPY wallet/api /app/api
 COPY wallet/public /app/public
 COPY wallet/official_assets/cape_v1_official_assets.lib /.espresso/verified_assets
 RUN chmod +x /app/wallet-api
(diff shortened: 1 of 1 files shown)


wallet.Dockerfile
```diff
@@ -4,6 +4,7 @@ RUN apt-get update \
   && rm -rf /var/lib/apt/lists/*
 
 COPY target/release/wallet-api /app/wallet-api
+COPY target/release/wallet-cli /app/wallet-cli
```
jbearer, line 7: I think we also need `RUN chmod +x /app/wallet-cli` (see below where we do it for `wallet-api`)

PR #1170, opened 2022-06-24
Get some parallelism in the faucet.

### LINKED ISSUE #1169: Faucet: break records in parallel
This looks good to me. It should increase our number of large records so we are not susceptible to a single large record get lost/delayed.

I think to get some parallelism and speed up the record breaking process, we need to do something about the `wallet.await_transaction(&receipt).await` in `break_up_records`. We should be able to skip this most of the time, since we expect to have other large records that can be broken up while the previous one is in transit. However, we don't want to skip this all the time -- if we did, then we might go back into the wait loop without creating enough records, in the case where we started with only one big record to break up. Therefore, I think we need to so something like
```
if wallet has more records that can be broken up {
  continue;
} else if this transaction will give us enough records {
  break;
} else {
  wait for transaction
}
```

Not sure if we want to do this now or in a separate PR, because this is already beneficial as is

_Originally posted by @jbearer in https://github.com/EspressoSystems/cape/pull/1167#pullrequestreview-1017397033_

jbearer:
Since the faucet now breaks records in half, creating more than one
big record at a time, we can generate multiple simultaneous transfers
when breaking up records, and then wait for them all at the same time,
rather than waiting for a transfer to complete before even building
the next one. Sometimes, we do not have to wait for the transfers to
complete at all.

This change also has the record breaker thread hold the wallet lock
less often, and simplifies the start-up process which waits for new
records to be created.

Closes #1169

### DIFF
--- a/faucet/src/faucet.rs
+++ b/faucet/src/faucet.rs
@@ -584,11 +584,11 @@ async fn spendable_records(
     })
 }
 
-/// Break large records into smaller records.
+/// Worker task to maintain at least `state.num_records` in the faucet wallet.
 ///
 /// When signalled on `wakeup`, this thread will break large records into small records of size
 /// `state.grant_size`, until there are at least `state.num_records` distinct records in the wallet.
-async fn break_up_records(state: FaucetState, mut wakeup: mpsc::Receiver<()>) {
+async fn maintain_enough_records(state: FaucetState, mut wakeup: mpsc::Receiver<()>) {
     loop {
         // Wait until we have few enough records that we need to break them up, and we have a big
         // enough record to break up.
@@ -626,11 +626,11 @@ async fn break_up_records(state: FaucetState, mut wakeup: mpsc::Receiver<()>) {
             wakeup.next().await;
         }
 
-        if let Some(transactions) = get_more_records(&state).await {
+        if let Some(transactions) = break_up_records(&state).await {
             // If we succeeded, wait until we are signalled again. Even though we may not have
             // enough records just yet, we will when the transactions submitted by
-            // `get_more_records` finalize. Returning to the previous loop and checking if we have
-            // enough records might spuriously lead us to call `get_more_records` again, which would
+            // `break_up_records` finalize. Returning to the previous loop and checking if we have
+            // enough records might spuriously lead us to call `break_up_records` again, which would
             // be an unnecessary waste of time.
             tracing::info!(
                 "will have sufficient records after {} transactions, waiting for a change",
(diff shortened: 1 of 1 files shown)


faucet/src/faucet.rs
```diff
@@ -616,7 +626,31 @@ async fn break_up_records(state: FaucetState, mut wakeup: mpsc::Receiver<()>) {
...
+        }
+    }
+}
+
+/// Break records into smaller pieces to create at least `state.num_records` total.
+///
+/// If successful, returns a list of transaction receipts which will give at least
+/// `state.num_records` when they are finalized. If there were not enough large records to break up
+/// to obtain the desired number of records, returns [None].
+async fn get_more_records(state: &FaucetState) -> Option<Vec<TransactionReceipt<CapeLedger>>> {
```
sveitser, line 649: I think this would call this function `break_up_records` and the existing `break_up_records` could be called `maintain_faucet_records` or something like that.

faucet/src/faucet.rs
```diff
@@ -626,30 +660,31 @@ async fn break_up_records(state: FaucetState, mut wakeup: mpsc::Receiver<()>) {
...
             }
 
             let largest_record = match records
                 .into_iter()
                 .max_by(|x, y| x.amount().cmp(&y.amount()))
             {
-                Some(record) => record,
-                None => {
-                    tracing::warn!("No spendable records");
+                Some(record) if record.amount() >= state.grant_size * 2u64 => record,
```
sveitser, line 679: Is the check on line 613 necessary if we have to check here as well?

sveitser (approved):
LGTM.

faucet/src/faucet.rs
```diff
@@ -504,28 +504,40 @@ async fn request_fee_assets(
...
-            );
-            let balance = wallet.balance(&AssetCode::native()).await;
-            let records = spendable_records(&wallet, state.grant_size).await.count();
-            tracing::info!(
-                "worker {}: wallet balance before transfer: {} across {} records",
-                id,
-                balance,
-                records
-            );
+            // If we don't have a sufficient balance, to transfer, it is probably only because some
```
philippecamacho, line 508: Is there not a risk to have this loop go on for ever? For example if the transactions in flight are rejected or if someone uses the same wallet somewhere else.

jbearer: I think it's okay. This worker task is supposed to live forever anyways. And as long as the balance is too low, it would not be able to make progress no matter what. If we allowed it to continue, it would only fail to transfer due to insufficient balance and send the failed request to the back of the queue.

Also, if the transactions in flight are rejected, we would get their inputs back, which would also allow this loop to exit.

faucet/src/faucet.rs
jbearer: The check on line 613 causes the thread to sleep if it fails. If we removed it then we would end up spamming `get_more_records` whenever we had no records to break up.

PR #1187, opened 2022-06-29
Use rolling log for EQS storage

### LINKED ISSUE #1183: Update CAPE + Seahorase to latest version of atomic store
[6:53 PM] Mat Richmond
Does the faucet do the write-to-disk thing like the address-book does?

[6:56 PM] 
Philippe Camacho added Jeb Bearer to the chat and shared all chat history.

[6:56 PM] Philippe Camacho
Mat RichmondDoes the faucet do the write-to-disk thing like the address-book does?
Any idea Jeb Bearer?

[6:58 PM] Jeb Bearer
In the healthcheck you mean? I don't believe it does

[6:59 PM] Mat Richmond
we should do that I think.

[6:59 PM] Mat Richmond
At the very least, it should (hopefully) cause healthchecks to fail and the service to restart.

[7:00 PM] Jeb Bearer
I'm not sure that's it, though. The faucet writes to disk anyways every time there is a request, and in the background threads when it generates a transaction

[7:00 PM] Jeb Bearer
I guess at least it would help us catch I/O errors

[7:00 PM] Jeb Bearer
I think the root problem more likely has to do with either running out of space or exceeding some inode or dirent limit because of how many small files it is creating

[7:05 PM] Philippe Camacho
Sorry for the naive question: why does the faucet need to create many files?

[7:07 PM] Mat Richmond
So for a service design, fail fast should be the preferred option. Right now we have failures but the health check continues to be good, which hopefully that would prevent.
 like 1

[7:08 PM] Mat Richmond
If the health check fails, then the service gets restarted and the symptoms are gone, even if the problem is still there. 

[7:08 PM] Mat Richmond
Which preserves service availability.

[7:09 PM] Jeb Bearer
Yep definitely support adding it to healthcheck for that reason

[7:10 PM] Jeb Bearer
Philippe Camacho Sorry for the naive question: why does the faucet need to create many files?
It does not. It uses an implementation of RollingLog from atomicstore that is supposed to keep snapshots of the most recent N states, but the current implementation never actually prunes old states. I think this might be fixed on the latest atomicstore? But not completely sure, NY should know

[7:10 PM] Jeb Bearer
If that is the case, I don't think upgrading CAPE to the latest atomicstore would be too hard
 heart 1 like 1

[7:10 PM] Jeb Bearer
CAPE + Seahorse, that is

[7:11 PM] Philippe Camacho
Cool, thanks Jeb Bearer

sveitser:
- Update atomic store to 0.1.3.
- EQS: keep 5 latest log entries.
- Return Error instead of panicking when commiting faucet grant index.
- [x] Use seahorse `0.2.6` when available.
- https://github.com/EspressoSystems/seahorse/pull/219

Should we make the number of log entries the EQS keep configurable via env var?

Close #1183

### DIFF
--- a/contracts/rust/Cargo.toml
+++ b/contracts/rust/Cargo.toml
@@ -63,7 +63,7 @@ rand = "0.8.4"
 rand_chacha = "0.3.1"
 reef = { git = "https://github.com/EspressoSystems/reef.git", tag = "0.2.2" }
 regex = "1.5.5"
-seahorse = { git = "https://github.com/EspressoSystems/seahorse.git", tag = "0.2.5" }
+seahorse = { git = "https://github.com/EspressoSystems/seahorse.git", branch = "mathis-atomicstore-0.1.3" }
 serde = { version = "1.0.124", features = ["derive"] }
 serde_json = "1.0.67"
 sha3 = "0.9.1"
--- a/contracts/rust/Cargo.toml
+++ b/contracts/rust/Cargo.toml
@@ -63,7 +63,7 @@ rand = "0.8.4"
 rand_chacha = "0.3.1"
 reef = { git = "https://github.com/EspressoSystems/reef.git", tag = "0.2.2" }
 regex = "1.5.5"
-seahorse = { git = "https://github.com/EspressoSystems/seahorse.git", branch = "mathis-atomicstore-0.1.3" }
+seahorse = { git = "https://github.com/EspressoSystems/seahorse.git", tag = "0.2.6" }
 serde = { version = "1.0.124", features = ["derive"] }
 serde_json = "1.0.67"
 sha3 = "0.9.1"
--- a/eqs/src/state_persistence.rs
+++ b/eqs/src/state_persistence.rs
@@ -7,16 +7,18 @@
 
 use crate::query_result_state::QueryResultState;
 use atomic_store::{
-    load_store::BincodeLoadStore, AppendLog, AtomicStore, AtomicStoreLoader, PersistenceError,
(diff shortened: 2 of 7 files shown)


contracts/rust/Cargo.toml
```diff
@@ -63,7 +63,7 @@ rand = "0.8.4"
 rand_chacha = "0.3.1"
 reef = { git = "https://github.com/EspressoSystems/reef.git", tag = "0.2.2" }
 regex = "1.5.5"
-seahorse = { git = "https://github.com/EspressoSystems/seahorse.git", tag = "0.2.5" }
+seahorse = { git = "https://github.com/EspressoSystems/seahorse.git", branch = "mathis-atomicstore-0.1.3" }
```
nyospe, line 66: Should this branch maybe be tagged instead?

eqs/src/state_persistence.rs
```diff
@@ -25,8 +27,10 @@ impl StatePersistence {
         store_path.push("eqs");
         let mut loader = AtomicStoreLoader::create(&store_path, key_tag)?;
         let snapshot_tag = format!("{}_state", key_tag);
-        let state_snapshot =
-            AppendLog::create(&mut loader, Default::default(), &snapshot_tag, 1024)?;
+        let mut state_snapshot =
+            RollingLog::create(&mut loader, Default::default(), &snapshot_tag, 1024)?;
+        state_snapshot.set_retained_entries(EQS_RETAINED_ENTRIES);
```
nyospe, line 32: 👍

contracts/rust/Cargo.toml
sveitser: I'm waiting for the seahorse PR to be merged, will update afterwards.

Issue #1179, opened 2022-06-28
`backend::test::test_anonymous_erc20_transfer` fails with `response body fails to deserialize`


sveitser:
I got some error when running the tests, but it might not be related to the change of this PR.

```
test backend::test::test_anonymous_erc20_transfer ... FAILED <69.927s>

failures:

---- backend::test::test_anonymous_erc20_transfer stdout ----
Sent funding tx to deployer
xfr: 987 bytes
mint: 971 bytes
freeze: 979 bytes
Sent funding tx to deployer
thread 'backend::test::test_anonymous_erc20_transfer' panicked at 'called `Result::unwrap()` on an `Err` value: Failed { msg: "relayer error: response body fails to deserialize: io error: unexpected end of file" }', wallet/src/backend.rs:995:14
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/result.rs:1749:5
   3: cape_wallet::backend::test::test_anonymous_erc20_transfer::{{closure}}
   4: std::thread::local::LocalKey<T>::with
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   6: async_io::driver::block_on
   7: std::thread::local::LocalKey<T>::with
   8: std::thread::local::LocalKey<T>::with
   9: async_std::task::builder::Builder::blocking
  10: core::ops::function::FnOnce::call_once
  11: core::ops::function::FnOnce::call_once
             at /rustc/7737e0b5c4103216d6fd8cf941b7ab9bdbaace7c/library/core/src/ops/function.rs:227:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    backend::test::test_anonymous_erc20_transfer

test result: FAILED. 14 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 69.93s

error: test failed, to rerun pass '-p cape_wallet --lib'
Sending HUP signal to run-geth: 14774
Cleaning up geth node
[...]

PR #1189, opened 2022-06-29
Check if storage is working when returning the response to a faucet healthcheck request.

### LINKED ISSUE #1186: Faucet: Check if storage is working when doing healthcheck

philippecamacho:
Closes #1186.

I am not sure if the storage requests are neutral.
Also I have an error when I run the tests locally. Something similar to the error described in #1179.

### DIFF
--- a/contracts/rust/src/bin/gen-vk-libraries.rs
+++ b/contracts/rust/src/bin/gen-vk-libraries.rs
@@ -18,7 +18,7 @@ use std::{fs::OpenOptions, io::prelude::*, path::PathBuf};
 
 // depth of the record merkle tree
 const TREE_DEPTH: u8 = 24;
-// list of supported tranaction types, each would result in a different verifying key
+// list of supported transaction types, each would result in a different verifying key
 const SUPPORTED_VKS: [(NoteType, u8, u8, u8); 7] = [
     (NoteType::Transfer, 1, 2, TREE_DEPTH),
     (NoteType::Transfer, 2, 2, TREE_DEPTH),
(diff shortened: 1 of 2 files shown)


sveitser:
I think this should do the job.

If only the remove steps fails (unlikely) then subsequent inserts would not actually write to disk anymore because if the key is already in the index it doesn't insert it again. This could be avoided by using a random key instead.

```rust
    /// Add an element to the persistent index.
    ///
    /// Returns `true` if the element was inserted or `false` if it was already in the index.
    fn insert(&mut self, key: UserPubKey) -> Result<bool, FaucetError> {
        if self.index.contains_key(&key) {
            // If the key is already in the index, we don't have to persist anything.
            return Ok(false);
        }
...
```

Another concern would be the faucet becoming unresponsive if the `/healthcheck` endpoint gets hit often because of the lock. This could be avoided by checking if we can write to disk in some other way (for example by creating a random file). The good thing about the current implementation is however that it exercises a lot more of the machinery we use in the faucet.

Overall I think it may reduce problems or alert us earlier so I'd be happy to try like this.

I'd be curious to get @jbearer's opinion though.

jbearer:
I don't think lock contention is a big problem. The time this healthcheck endpoint spends with the lock is nothing compared to how long it takes to build a transaction. Plus the healthcheck is only called every few seconds, I believe.

If only the remove step fails, then we will actually end up sending a grant to the default address, which wastes a small amount of funds and time, but it's not too bad given that this should be extremely rare. And when we send the grant to the default address, we will remove it from the queue, so subsequent healthchecks will start doing the correct thing again.

Only suggestion is I think we should log something at ERROR level if either of these writes fails, so that if the healthcheck fails we will know why, and especially if only one of the writes fails, we will know it happened.

sveitser (approved):
LGTM

PR #1192, opened 2022-07-13
Drop write lock before persisting to disk

sveitser:
Also adds a few info logs related to EQS activity.

I think this is preferable to operating on a copy because all operations on the state go through the RWLock directly. In my experiments this reduced the lock time from up 1 second to about 40ms. The less safe alternative (making a copy and then acquiring a write lock to write back the modified copy) locks for about 20ms. However with the second version it would be possible to operate on more than one copy simultaneously, thereby ending up with a wrong state.


philippecamacho (approved):
LGTM

PR #1195, opened 2022-07-14
Persist EQS state after loop over events

sveitser:
This should reduce the number of writes to storage, especially if the EQS
has falled behinds and fetches multiple transactions at once.

The downside is that the EQS would have to reprocess events if an error
occurs during the loop.

The EQS log level is set to WARN because of tide-rs. As a workaround use
WARN for for EQS INFO logs.


nyospe (approved):
An acceptable workaround. Will speed up initial load a lot, and will allow faster catch up when EQS falls behind.

PR #1196, opened 2022-07-14
Undo WARN log workaround

sveitser:
We can configure the logs the way we want with

    RUST_LOG="info,tide=warn,net=warn"


VictorKoenders (approved):

VictorKoenders:
You could also do it the other way around, `RUST_LOG="warn,eqs=info"`. Either way works!

PR #1207, opened 2022-08-08
Rejects when the submitted block and the pending deposits queue are both empty

### LINKED ISSUE #1206: Do not process empty blocks without pending deposits in the CAPE contract
See #1203 and this [discussion](https://espresso.zulipchat.com/#narrow/stream/311688-CAPE/topic/Future.20of.20CAPE/near/292042846) for context.

philippecamacho:
Closes #1206

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 .idea
 node_modules
 cache
-.env
+.env*
 **/*.DS_Store
 .log/
 
--- a/contracts/contracts/CAPE.sol
+++ b/contracts/contracts/CAPE.sol
@@ -282,6 +282,9 @@ contract CAPE is RootStore, AssetRegistry, ReentrancyGuard {
         uint256 freezeIdx = 0;
         uint256 burnIdx = 0;
 
+        // We require either the block or the pending deposits queue to be non empty.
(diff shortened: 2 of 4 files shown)


sveitser:
LGTM but I wonder why not checking if there are no commitments instead of checking if there no pending deposits and no notes? It seems a bit simpler to me and since commitments are what go into the merkle tree I think it easier to understand why we reject the block.

In any event I think it would be good to add a comment somewhere about why we reject the block and what problem is caused by not doing that.

sveitser (changes requested):
I think we should add a comment as to why we reject empty blocks.

philippecamacho:
> LGTM but I wonder why don't we check if there are no commitments instead of checking if there no pending deposits and no notes? It seems a bit simpler to me and since commitments are what go into the merkle tree I think it easier to understand why we reject the block.
> 
Yes, I agree. On the other side it seems to save a bit of gas to reject the block right away. WDYT?

> In any event I think it would be good to add a comment somewhere about why we reject the block and what problem is caused by not doing that.

Yes, added in e963fc1

philippecamacho:
> I think we should add a comment as to why we reject empty blocks.

Yes, added in e963fc1.

sveitser (approved):
LGTM

PR #1210, opened 2022-08-16
Use rust 1.63.0

sveitser:
The issue with LLVM is fixed in stable now.

https://github.com/rust-lang/llvm-project/commit/fc0ffca0ba2f6bc479433ff52634deeb764c2585


philippecamacho (approved):
LGTM

PR #1228, opened 2022-10-27
Use EIP1559 transactions

sveitser:
Some tests are failing. They can likely be fixed by manually specifying a high enough gas limit.


philippecamacho:
Any reason why we should not fix the tests, merge to main and deploy the changes to the Goerli services?

sveitser:
The only reason would be that it's not enough of a priority to make this change.

PR #1264, opened 2023-01-11
Update rust 1.63.0 -> 1.66.1

sveitser:
There would be quite a few lints to fix to get clippy to pass with `-D warnings`.


nyospe (approved):
Good enough for now. Best to come back and fix the lints later.

shenkeyao (approved):
LGTM

sveitser:
@Ancient123 it looks like the arm64 build job isn't picked up by any runners: https://github.com/EspressoSystems/cape/actions/runs/3894058849/jobs/6653143567

